PR #132, opened 2026-02-28
Add CodeQL analysis for all 5 SDK languages

jeremy:
## Summary

- Adds CodeQL interprocedural taint-tracking analysis across Go, TypeScript, Ruby, Swift, and Kotlin
- Path-filtered: on PRs, only scans languages whose SDK directories changed; push-to-main/schedule/dispatch scans all
- Matrix workflow for Go/JS/Ruby/Kotlin on `ubuntu-latest`; Swift in a separate job on `macos-15`
- Generated code excluded via two layers: `paths-ignore` for source extraction (JS/Ruby) and a SARIF post-filter step for compiled languages where build tracing captures generated code regardless
- Analysis/upload separated so extractor errors break the build while GHAS-unavailable upload errors are tolerated
- Fixture-based regression test for the SARIF filter jq expression (relative URIs, absolute URIs, null/missing `.results`)

## Test plan

- [x] Workflow triggers on this PR with all 5 matrix jobs visible in Actions
- [x] Each job completes analysis (succeeds even without GHAS)
- [x] Swift job runs on macOS, all others on Ubuntu
- [x] Swift job skipped when PR doesn't touch `swift/**`
- [x] Config changes (`.github/codeql/**`, `.github/workflows/codeql.yml`) trigger all languages
- [x] Filter step logs `filtered N generated-code alerts` counts per SARIF file
- [x] SARIF uploads succeed for all 5 languages (GHAS default setup disabled)
- [x] `make check` passes (verified locally)
- [x] `.github/codeql/test-sarif-filter.sh` passes locally

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -18,31 +18,68 @@ jobs:
     name: Detect changes
     runs-on: ubuntu-latest
     outputs:
-      swift: ${{ github.event_name != 'pull_request' || steps.filter.outputs.swift == 'true' }}
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+      swift: ${{ steps.set-matrix.outputs.swift }}
     steps:
       - uses: dorny/paths-filter@v3
         if: github.event_name == 'pull_request'
         id: filter
         with:
           filters: |
+            go:
+              - 'go/**'
+            typescript:
+              - 'typescript/**'
+            ruby:
+              - 'ruby/**'
+            kotlin:
+              - 'kotlin/**'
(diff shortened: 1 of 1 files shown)


jeremy:
**Re: unpinned `dorny/paths-filter@v3`** — Same pattern as all other actions in this repo (tag refs, not pinned hashes). Can pin across all workflows in a dedicated follow-up.

**GHAS upload now working** — default setup was disabled. All 5 languages uploaded successfully: `Successfully uploaded results` in every job. Security tab should now show findings per `codeql-*` category.

PR #291, opened 2026-04-28
Add bucketId to card column color and on-hold paths

packagethief:
## Summary

Three card column operations had bucket-less URIs in `spec/basecamp.smithy`, so the generated clients in all six SDKs built paths like `/{accountId}/card_tables/columns/{columnId}/color.json`. Basecamp's API rejects those with 404 — the endpoints have always required `/buckets/{bucketId}` in the path. **Every consumer of `SetCardColumnColor`, `EnableCardColumnOnHold`, and `DisableCardColumnOnHold` has been broken since these operations were added.**

Reported via [Basecamp card 9825845889](https://3.basecamp.com/2914079/buckets/27/card_tables/cards/9825845889) (Helpscout case from a Jack Henry customer). The basecamp-cli has a stopgap workaround at basecamp/basecamp-cli#455 — once this PR ships and CLI bumps to the new SDK, that workaround will be reverted.

## Reproduction (verified live)

- `PUT /3642500/card_tables/columns/7103332776/color.json` → **404 Not Found**
- `PUT /3642500/buckets/36395297/card_tables/columns/7103332776/color.json` → **200 OK**

## Spec change

Added `@httpLabel bucketId: ProjectId` to all three operation inputs and rewrote each `@http` URI to include `/buckets/{bucketId}/`, matching the pattern used by other bucket-scoped operations (`ListWebhooks`, etc.).

`GetCardColumn` and `UpdateCardColumn` paths are **not** changed in this PR. Those happen to work without the bucket today (Basecamp's API tolerates bucket-less reads on these specific endpoints), and changing them would be a larger breaking change with no functional benefit. Worth a follow-up to re-align them with the spec for consistency.

## Breaking change

Go SDK: `SetColor`, `EnableOnHold`, `DisableOnHold` now take a `bucketID int64` parameter before `columnID`. TypeScript / Ruby / Python / Swift / Kotlin generated method signatures gain a `bucketId` parameter. Since the previous signatures couldn't ever produce a successful API call, no currently-working caller is affected.

## Regeneration

Ran the full pipeline:
- `make smithy-build` (Smithy → openapi.json)
- `make -C go generate && make url-routes`
- `make ts-generate ts-generate-services`
- `make rb-generate rb-generate-services`
- `make swift-generate`
- `make kt-generate-services`
- `make py-generate`

Hand-written updates:
- `go/pkg/basecamp/cards.go` — `SetColor` / `EnableOnHold` / `DisableOnHold` wrappers take `bucketID` and pass it through.
- `typescript/tests/services/card-columns.test.ts` — call sites and MSW stubs updated.
- `ruby/test/basecamp/services/card_columns_service_test.rb` — call sites and WebMock stubs updated.
- No Python / Swift / Kotlin tests reference these methods today.

## Test plan

- [x] `make smithy-check` — spec round-trips cleanly.
- [x] `make go-check-drift` / `kt-check-drift` — no service-layer drift.
- [x] `make ts-check` — 616 tests passed.
- [x] `make rb-check` — 599 runs / 905 assertions passed; rubocop clean.
- [x] `make py-check` — 209 tests passed; mypy + ruff clean.
- [x] `make kt-check` / `swift-check` — pass.
- [x] `make conformance` — 64 passed, 0 failed.
- [x] `cd go && make test` — pass. (`make go-check` skipped: golangci-lint not installed locally; CI will exercise it.)
- [ ] CI green.

## Follow-ups (not blocking this PR)

1. Revert the basecamp-cli workaround in basecamp/basecamp-cli#455 after this ships and the CLI bumps the SDK.
2. Consider aligning `GetCardColumn` and `UpdateCardColumn` paths to be bucket-scoped for spec consistency.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jeremy (approved):
Approve. Spec fix is correct: cross-checked against the BC3 Rails app — those controllers are `BucketScoped` and the routes live under `resources :buckets`, so `/buckets/{bucketId}/...` is the right target. Some published API doc examples show the bucket-less form, but those are stale; the Rails routes are the source of truth and reject the bucket-less form with 404 (reproduced).

Hand-written `cards.go` wrapper signature changes (`SetColor`, `EnableOnHold`, `DisableOnHold`) and the TS/Ruby test stub updates are correct. Regen is consistent across all six SDKs. The `go.work.sum` diff isn't unrelated churn — running `go build ./...` from the parent reproducibly re-adds the same 5 lines, so this PR is correcting pre-existing workspace-sum drift on `main`. @cubic found no issues on the original commit.

Added two fixup commits onto the branch:

1. **`9c15b67` Add CardColumns httptest coverage for bucket-scoped paths.** TS and Ruby already assert URL paths in their test stub diffs; Go didn't. Adds three `httptest`-backed tests pinning method and full path for `SetColor` / `EnableOnHold` / `DisableOnHold` using distinct `bucketID` and `columnID` constants — a future swap of the wrapper's `(bucketID, columnID)` argument order builds the path with the IDs in the wrong slots and fails the test, rather than silently producing a 404 against the live API like the bug this PR fixes.

2. **`58c00e3` Model RateLimitError on retry-marked card column ops.** Picks up @copilot's two flags. `EnableCardColumnOnHold` already declared `RateLimitError`; aligning `SetCardColumnColor` and `DisableCardColumnOnHold` so all three siblings (which share `@basecampRetry(retryOn: [429, 503])`) model the 429 response consistently. Spec edit + clean regen across the affected artifacts. The wider asymmetry (203 retry-marked ops vs 97 modeling `RateLimitError`) is a separate cleanup PR.

CI is green on head `58c00e3`.

One follow-up worth filing separately: `grep` for these three operations in `conformance/tests/*.json` returns nothing, which is exactly how a wrong-path bug reaches the live API undetected. Adding conformance fixtures (templated on the recent uploads-download work in #280/#281) would close the gap for this class of regression.

PR #442, opened 2026-07-25
CI: add Swift generated-code drift gate

jeremy:
## What

Add a generated-code drift gate for Swift, closing the gap that let a stale
`Metadata.swift` reach `main` in #440.

`make check` already runs drift gates for Go (`go-check-drift`), Kotlin
(`kt-check-drift`), and Python (`py-check-drift`) — but none for Swift/TS/Ruby.
Swift is a clean regen-and-diff candidate: everything under
`swift/Sources/Basecamp/Generated/` is generator-emitted (`Models/` and
`Services/` are wiped+rewritten; the root files `AccountClient+Services.swift`,
`Metadata.swift`, `QueryString.swift` are each written directly) and embeds
**no timestamp**, so a plain diff against a fresh regen is exact.

## Changes

- **`scripts/check-swift-service-drift.sh`** (new, `100755`): mirrors
  `check-python-service-drift.sh`. Regenerates into a `mktemp -d` (invoking the
  generator directly, not `make swift-generate`, which hardcodes the committed
  output path), then diffs against the committed tree. The diff is guarded so
  `set -e` can't swallow the diagnostic before the error message prints.
- **`Makefile`**: new `swift-check-drift` target guarded exactly like
  `swift-check` (`ifdef IS_MACOS … else SKIP`); added to `.PHONY`, to the
  `check` dependency list next to `kt-check-drift`, and to help text. On Linux
  it prints SKIP; on macOS it runs.
- **`.github/workflows/test.yml`**: a "Check generated code drift" step in the
  `test-swift` job (`macos-15`, Swift toolchain already present), so the gate
  has real teeth in CI.

## Why enhancement

Tooling/CI improvement — no wire or behavior change.

## Verified

- Clean tree → script exits 0; hand-edited generated file → exits 1 with a clear
  diagnostic; reverted → 0 again.
- `make check` green (gate runs on macOS); `make swift-check-drift IS_MACOS=`
  prints SKIP and exits 0 (Linux path).
- `make lint-actions` (actionlint + zizmor) clean.

## Follow-up

TS `metadata.ts` and Ruby `metadata.json`/`types.rb` embed a `generated:`
timestamp; a drift gate for those needs timestamp normalization first, left out
of scope here.

---
## Summary by cubic
Add a Swift generated-code drift gate to block stale files from merging. Runs in CI and via `make check` to ensure `swift/Sources/Basecamp/Generated` matches a fresh regen.

- **New Features**
  - Adds `scripts/check-swift-service-drift.sh` to regen into a temp dir and diff against committed output.
  - Adds `swift-check-drift` Makefile target; included in `check` and gated to macOS (prints SKIP on non-macOS).
  - Adds a "Check generated code drift" step to the `test-swift` GitHub Actions job.

- **Migration**
  - If the drift check fails, run `make swift-generate` and commit the changes.

<sup>Written for commit cafb5bbf787d429ea08e6b3895655473cbbfe52d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/basecamp/basecamp-sdk/pull/442?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -275,6 +275,9 @@ jobs:
 
       - name: Test
         run: swift test
+
+      - name: Check generated code drift
+        run: ../scripts/check-swift-service-drift.sh
   test-kotlin:
     name: Kotlin Tests
     runs-on: ubuntu-latest
--- a/Makefile
+++ b/Makefile
@@ -623,7 +623,7 @@ gradle-stop:
 HAS_SWIFT := $(shell command -v swift 2>/dev/null)
 IS_MACOS  := $(filter Darwin,$(shell uname -s))
 
-.PHONY: swift-build swift-test swift-check swift-clean swift-generate
+.PHONY: swift-build swift-test swift-check swift-check-drift swift-clean swift-generate
 
 # Build Swift SDK (macOS only — SDK requires Apple platforms)
 swift-build:
@@ -647,6 +647,14 @@ ifdef IS_MACOS
 	@$(MAKE) -C swift check
 else
 	@echo "SKIP: swift-check (macOS only)"
+endif
+
+# Check for drift between generated Swift code and the OpenAPI spec (macOS only)
+swift-check-drift:
+ifdef IS_MACOS
+	@scripts/check-swift-service-drift.sh
+else
+	@echo "SKIP: swift-check-drift (macOS only)"
 endif
 
 # Regenerate Swift SDK services from OpenAPI spec (needs swift on any platform)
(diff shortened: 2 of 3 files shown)


jeremy:
Superseded by #444, which adds the same Swift generated-drift gate (`scripts/check-swift-service-drift.sh` + `make swift-check-drift` + CI step) as part of the idempotency-classification hardening series (#444 → #445 → #446).

One deliberate difference: #444 gates `swift-check-drift` on **`HAS_SWIFT`** rather than `IS_MACOS`. Generating the Swift sources needs only the Swift toolchain (not an Apple platform), so the freshness gate runs on any OS that has `swift` — including the Ubuntu runners — whereas `IS_MACOS` would have skipped it everywhere except the macOS job. Closing this in favor of #444.

Issue #480, opened 2026-07-28
Two operations under-classified as non-idempotent: Subscribe and EnableCardColumnOnHold


jeremy:
`@basecampIdempotent(natural: true)` currently covers 69 operations. Two more meet the same bar and are missing it. Both are conservative gaps — under-classification costs a retry, it does not risk a duplicate — but they falsify the claim that the 69 are complete, and they make two pairs of sibling operations inconsistent with each other for no reason anyone recorded.

Verified against BC3 `origin/master` (`2f01d0da`), route-resolved through `config/routes.rb`.

## 1. `Subscribe` — identical implementation to the flagged `SubscribeToCardColumn`

| | `Subscribe` | `SubscribeToCardColumn` |
|---|---|---|
| Route | `POST /{accountId}/recordings/{recordingId}/subscription.json` | `POST /{accountId}/card_tables/lists/{columnId}/subscription.json` |
| Controller | `Recordings::SubscriptionsController#create` | `Kanban::Lists::SubscriptionsController#create` |
| Body | `@recording.subscribe Current.person, track: true` | `@list.subscribe Current.person, track: true` |
| Smithy | `spec/basecamp.smithy:5381` — **unflagged**, `maxAttempts: 2` | `spec/basecamp.smithy:4595` — `@basecampIdempotent(natural: true)`, `maxAttempts: 3` |

Both reach the same `Recording::Subscribable#subscribe`, which is idempotent by construction (`app/models/recording/subscribable.rb:20`):

```ruby
def create_subscriptions_for_eligible(people)
  (eligible_as_subscribers(people) - subscribers).select do |person|
    begin
      subscriptions.create!(subscriber: person)
    rescue ActiveRecord::RecordNotUnique
      false
    end
  end
end
```

Already-subscribed people are subtracted out, a race is rescued as `RecordNotUnique`, and the surrounding `if people.any?` suppresses tracking when nothing changed — so a repeat POST creates no duplicate subscription and emits no duplicate event.

## 2. `EnableCardColumnOnHold` — guarded create, and its DELETE sibling is already flagged

| | `EnableCardColumnOnHold` | `DisableCardColumnOnHold` |
|---|---|---|
| Route | `POST .../card_tables/columns/{columnId}/on_hold.json` | `DELETE` on the same path |
[...]

Issue #479, opened 2026-07-28
Retry metadata fidelity: Go and Python ignore per-operation retry_on and max_attempts


jeremy:
Gates 1 and 2 of the §7 three-gate algorithm read `behavior-model.json`. Gate 3's *parameters* do not — Go and Python ignore the per-operation `retry` block entirely, so declared retry policy and actual retry behavior have drifted apart.

Idempotency classification is pinned by `check-idempotency-parity`. Retry **status set** and **attempt count** are not pinned by anything.

## What is declared

All 226 operations declare `retry_on: [429, 503]`. `max` is `2` for 43 operations (9 of them idempotent) and `3` for the other 183.

## What actually happens

| | Declared | Actual |
|---|---|---|
| Statuses | `retry_on: [429, 503]` per operation | **Go**: global `isRetryableStatus` → `{429, 500, 502, 503, 504}` (`go/pkg/generated/client.gen.go:3963`). **Python**: no status gate at all — `python/src/basecamp/errors.py:153-156` marks 500 and 502/503/504 `retryable=True`, and the retry loop keys off `e.retryable` |
| Attempts | per-operation `max` | **Go**: global `RetryConfig.effectiveMaxAttempts()` (`client.gen.go:3982`); `OperationMetadata` (`client.gen.go:19572`) carries **no attempt count at all**. **Python**: global `Config.max_retries` (`_http.py:152`, and the same in `_async_http.py`) |

Net effect: an operation declaring `max: 2` is attempted three times under default config, and a 500 triggers retries that `retry_on` never sanctioned.

## Fix

**Statuses** — honor the operation's declared `retry_on` on HTTP-response retry paths *that carry behavior-model metadata*. Go: replace the global `isRetryableStatus` with the operation's declared set. Python: gate on the declared statuses **without** changing what `errors.py` reports to callers (`retryable` stays part of the public error surface).

**Attempts** — normalize first; the operation max is a **ceiling**, never a floor:

```
caller_cap         = max(1, configured_attempts)
effective_attempts = min(caller_cap, operation_max)
```

`Config(max_retries=0 or 1)` must remain **one attempt** (§2 documents 0 as an accepted compatibility value). Go needs a new attempt field on `OperationMetadata`; Python must apply the per-operation ceiling.

### Python: four call sites, not one

| | Transport | Generated base |
|---|---|---|
| Sync | `python/src/basecamp/_http.py` | `generated/services/_base.py:48` |
[...]

PR #484, opened 2026-07-28
Retry docs: correct the mutation and 401 claims, add Python to the roster

jeremy:
Documentation only — no behavior change. Every claim below was re-derived from source at `d5604962`.

## `SECURITY.md` — two false bullets

**"Mutations (POST/PUT/DELETE): NOT automatically retried on 429/503".** They are. `behavior-model.json` flags 69 operations `idempotent: true` — all 45 PUTs, all 21 DELETEs, and 3 POSTs (`CompleteTodo`, `PauseQuestion`, `SubscribeToCardColumn`) — and those retry in Go's generated operation path, Python, TypeScript, Kotlin, and Swift. **Ruby** is the only transport that is GET-only (`http.rb` routes every non-GET to `single_request`), as is Go's separate hand-written `pkg/basecamp` helper. The bullet dates to #60, before the generated client existed.

The 39 non-idempotent POSTs are single-attempt everywhere, which is the property the bullet was reaching for — so the replacement states the gate as idempotency rather than "is it a mutation".

**"401 responses: Token refresh attempted, then single retry for all methods".** True in three transports, false in four:

| SDK | 401 behavior |
|-----|--------------|
| Go — hand-written `pkg/basecamp` | Refresh via `AuthManager`, then one retry (`client.go:794-800`) |
| Go — generated `pkg/generated` | **No reactive refresh.** `AuthTransport.RoundTrip` fetches a token proactively (`client.gen.go:19851`); `ParseHTTPError` turns 401 into `AuthError` (`:20237`) |
| Ruby | Refresh, then one retry (`http.rb:417`) |
| Python (sync **and** async) | Refresh, then one retry (`_http.py:237`, `_async_http.py:237`) |
| TypeScript / Kotlin / Swift | No refresh; the 401 reaches the caller |

Swift's `HTTPClient.swift:157` re-runs the auth strategy before each *retry*, but 401 is not a retryable status, so that path never fires for a 401.

## `SPEC.md` — Python was missing, Ruby was miscategorized

- **`:11`** "Five shipping SDKs" and **`:38`** "consensus of Go, Ruby, TypeScript, Kotlin, Swift" both omitted Python. Six now.
- **`:401`** named "Go, Ruby, and Swift" as retrying network errors. Python belongs there (`NetworkError` is `retryable=True` and the loop keys off it). Ruby does **not** belong in the same group — it retries network errors only on GET, because `request` sends every non-GET to `single_request`. Listing it alongside Go/Swift implied a parity it does not have.
- **Appendix F** had no Python row.

## New: Gate 3 fidelity note

Gates 1 and 2 read `behavior-model.json`. Gate 3's *parameters* do not, in two SDKs — Go and Python substitute a client-global attempt count for the per-operation `retry.max` and treat a superset of `retry_on` as retryable. Go's `OperationMetadata` carries no attempt count at all. That is now stated in §7 and footnoted under Appendix F, and filed as **#479** with the fix design.

Also filed **#480**: `Subscribe` and `EnableCardColumnOnHold` are under-classified as non-idempotent (`Subscribe` calls the identical `Recording::Subscribable#subscribe` that the flagged `SubscribeToCardColumn` calls; `EnableCardColumnOnHold`'s create is `unless @on_hold`-guarded and its DELETE sibling is already flagged). Both verified against BC3 `2f01d0da`.

## Deliberately left alone

- **`SPEC.md:400`** — *"Implementations may expand this set to include other 5xx statuses"* sanctions the Go/Python behavior. It should go, but only together with the #479 fix; removing it now would make the spec false in the other direction.
- **`SPEC.md:407` / `:411` / `:1591` and waiver 2B.1** — the TypeScript one-retry cap is **current, not stale**: `client.ts:1007` returns native `fetch(retryRequest)`, bypassing middleware, so the `X-Retry-Attempt` header it sets is never re-consumed. `conformance/runner/typescript/runner.test.ts:73-78` pins it.

`make check` passes.

---
## Summary by cubic
Updates retry docs to match reality: idempotent mutations do retry on 429/503, 401 handling varies by SDK, and Python is now included. Also clarifies network-error retry behavior, per-transport 401 differences, Ruby’s raw-upload exception, and Gate 3 fidelity.

- **Bug Fixes**
  - SECURITY.md: Replace “mutations never retry” with idempotency rules; Ruby and Go’s hand-written helper are GET-only; respect Retry-After; non-idempotent POSTs never retry on 429/503 or network errors.
  - SECURITY.md: Add per-SDK 401 matrix — reactive refresh only in hand-written Go, Ruby, and Python; generated Go `*WithResponse` fills `JSON401` and returns no error; TypeScript service wrappers throw on 401 but the raw client returns `{ data, error }` without throwing; Kotlin/Swift surface 401; Ruby raw uploads (`post_raw`/`put_raw`) do not replay.

- **Spec Updates**
  - Add Python across the spec and Appendix F; update “five” to “six”; fix network-error grouping: Go/Python/Swift retry; Ruby GET-only; TypeScript retries only after a response; Kotlin doesn’t retry network errors; clarify OAuth discovery SSRF note (Swift out of scope).
  - Gate 3 fidelity: Go/Ruby retry a superset of `retry_on`; Python has no status gate (uses a `retryable` flag); TypeScript/Kotlin/Swift use the declared set. Ruby ignores per-operation retry ceilings; others honor them.

<sup>Written for commit 8af46a1f4f603b4ca0085655fd24b184b368c932. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/basecamp/basecamp-sdk/pull/484?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -183,12 +183,38 @@ print(f"Headers: {safe}")
 
 ## Retry Behavior
 
-The SDKs implement safe retry behavior:
+Retry eligibility is decided per *operation*, not per HTTP method. `behavior-model.json` classifies
+all 226 operations: the 118 GETs are retryable by method, and 69 mutations are flagged
+`idempotent: true` — all 45 PUTs, all 21 DELETEs, and 3 POSTs (`CompleteTodo`, `PauseQuestion`,
+`SubscribeToCardColumn`). The other 39 POSTs are attempted exactly once. SPEC.md §7 specifies the
+three-gate algorithm and the per-SDK divergences.
 
-- **GET requests**: Automatically retried with exponential backoff on 429/503
-- **Mutations (POST/PUT/DELETE)**: NOT automatically retried on 429/503 to prevent duplicate operations
-- **401 responses**: Token refresh attempted, then single retry for all methods
-- **Retry-After headers**: Respected for 429 responses
+- **Reads (GET)**: retried with exponential backoff on 429/503 in every SDK. (HEAD is idempotent by method too, but Ruby's transport gates on `method == :get` specifically, so a HEAD would not retry there. The API surface has no HEAD operations today, so this is theoretical.)
+- **Naturally-idempotent mutations (PUT/DELETE) and the 3 flagged POSTs**: *are* retried on 429/503
+  by Go (generated operation path), Python, TypeScript, Kotlin, and Swift. Retrying these cannot
+  duplicate a resource, which is why the gate is idempotency rather than "is it a mutation".
+  **Ruby is the sole exception** — its transport retries GET only.
+  Go's separate hand-written `pkg/basecamp` HTTP helper is also GET-only.
+- **Non-idempotent POSTs**: never retried on 429/503 or network failure, in any SDK — a retry could create a duplicate resource. This does **not** mean such a POST is always attempted exactly once: a 401 that triggers a successful token refresh replays the request once regardless of idempotency, in Ruby and both Python transports (see the 401 table below). Ruby's **raw upload** path is the exception — `post_raw`/`put_raw` (attachments, campfire uploads) go through `single_request_raw`, which raises the mapped error directly and has no refresh-and-replay branch, so those POSTs really are attempted exactly once.
+- **Retry-After headers**: respected for 429 responses.
+
+### 401 handling is not uniform
+
+Reactive token refresh — refresh on a 401, then retry the request once — exists in only three
+transports. Everywhere else the 401 is surfaced to the caller:
+
+| SDK | 401 behavior |
+|-----|--------------|
+| Go — hand-written `pkg/basecamp` | Refresh via `AuthManager`, then a single retry (`client.go`) |
+| Go — generated `pkg/generated` operations | **No reactive refresh.** `AuthTransport` obtains a token proactively per request via `TokenProvider.AccessToken`. A 401 is **not** returned as an error by the `*WithResponse` variants: they populate `response.JSON401` and return `(response, nil)`, so a caller checking only `err` would read the request as successful. Check `JSON401` (or the status code) explicitly. The `ParseHTTPError` helper is what maps a 401 to `AuthError` |
+| Ruby | Refresh, then a single retry (`http.rb`) — **except** the raw upload path: `single_request_raw` (used by `post_raw`/`put_raw` for attachments and campfire uploads) raises without replaying |
+| Python (sync **and** async) | Refresh, then a single retry (`_http.py`, `_async_http.py`) |
+| TypeScript | No refresh. Generated **service wrappers** convert a 401 into a thrown `BasecampError`; the **raw client** does not — `BasecampClient` extends `RawClient`, so `client.GET(...)` resolves with `{ data: undefined, error }` and never throws. A caller using the raw API with `try`/`catch` alone will miss authentication failures |
+| Kotlin | No refresh; raised as `BasecampException.Auth` |
+| Swift | No refresh; raised as `BasecampError.auth`. The transport re-runs its auth strategy before each *retry*, but 401 is not a retryable status, so that path is never reached for a 401 |
+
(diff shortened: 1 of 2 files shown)


jeremy:
Rebased onto current #483 (`aee8bd98` + the rationale fix). It was based on the superseded `7a4c3297`, which is why it went CONFLICTING and exposed 55 files. Back to **2 files**, `SECURITY.md` and `SPEC.md`.

All six review findings applied — each verified against source first:

1. **`SPEC.md:40` "all 5 SDKs use 500"** — correct, and it wasn't the only one. Fixed line 40, the matching `[CONFLICT]` note at :658, the SSRF heading, and six rows of the constants table. Python does share every one of them (`errors.py:186` `_truncate(max_bytes=500)`, and `config.py:21-25` for the defaults), so "six" is accurate rather than just arithmetic.

2. **GET/HEAD is not uniform** — correct. Ruby gates on `method == :get` specifically, so a HEAD would not retry there. Narrowed to GET, with the divergence noted and the fact that there are no HEAD operations in the surface today.

3. **The single-attempt POST claim needed qualifying** — correct, and my own table two paragraphs below already contradicted it. Now scoped: never retried *on 429/503 or network failure*, with an explicit pointer to the 401-refresh replay in Ruby and both Python transports.

4. **Generated Go 401s are typed responses** — correct and the sharper of the two Go findings. `*WithResponse` populates `response.JSON401` and returns `(response, nil)`; a caller checking only `err` reads a 401 as success. The table now says that, and notes `ParseHTTPError` is the helper that maps 401 → `AuthError`.

5. **Ruby doesn't honor the per-operation ceiling** — correct. The Appendix F footnote promised a universal ceiling. It now names the three tiers: Go and Python honor it *and* let a caller ask for fewer; TypeScript/Swift/Kotlin honor it but override a lower caller cap where one exists; Ruby never consults it.

6. **The cubic P1 on `isRetryableStatus` is wrong** — I checked `client.gen.go` at this base rather than reasoning about it. `isRetryableStatus` really is `{429, 500, 502, 503, 504}` (`client.gen.go:3964-3974`). The comparison to `checkResponse`'s `>= 500 && < 600` in `helpers.go` is a different function on the hand-written path; it does not govern generated-operation retry. The doc's narrower set is accurate, so no change. #486 is what replaces that global set with the per-operation declared one.

jeremy:
Three more corrections, all verified against source first.

**Ruby is excluded from the per-operation ceiling claim.** The Appendix F footnote said the ceiling is "honored in every SDK". `request_with_retry` loops until `@config.max_retries` (`ruby/lib/basecamp/http.rb`) without reading operation metadata at all, so a Ruby caller who raises the cap above 3 exceeds the operation's declared `max`. Now names Ruby explicitly, and notes the separate Kotlin direction (honors the ceiling but overrides a *lower* caller cap) with a pointer to #485.

**Ruby's 401 replay excludes raw uploads.** `post_raw`/`put_raw` — attachments and campfire uploads — go through `single_request_raw`, whose rescue raises the mapped error directly with no refresh-and-replay branch (`http.rb:461-471`), unlike `single_request`. So for those POSTs the "attempted exactly once" claim I had just qualified away is in fact true. Both the bullet and the Ruby row in the 401 table now carry the exception.

**TypeScript's 401 depends on which API you call.** `BasecampClient extends RawClient`, so `client.GET(...)` resolves with `{ data: undefined, error }` and never throws — pinned by `typescript/tests/client.test.ts:335-353`. Only the generated service wrappers convert that into a thrown `BasecampError`. The table row said flatly "raised to the caller", which would lead someone using the raw API with `try`/`catch` to miss authentication failures entirely. This is the same shape as the generated-Go typed-401 finding fixed earlier, and it now reads the same way.

PR #487, opened 2026-07-28
Search: model content/description as required and nullable, add the plain-text excerpts

### LINKED ISSUE #465: SearchResult.content/description are always null on the wire; plain_text_content/plain_text_description are unmodeled
## Summary

`SearchResult.content` and `SearchResult.description` are modeled and typed in all six SDKs, but **BC3 always sends them as `null`**. The fields that actually carry the searchable text — `plain_text_content` and `plain_text_description` — are **not modeled at all**.

Net effect: an SDK consumer running a search gets a result whose text-bearing fields are typed, present, and permanently empty, with no typed access to the text BC3 does return.

Surfaced while verifying #428 (see #464); filed separately because it is a distinct and larger gap.

## Evidence — bc3 `c308693171` (the SHA `spec/api-provenance.json` pins)

`app/views/api/searches/show.json.jbuilder`:

```ruby
json.partial! api_search_result_template_path(recording), recording: recording

# Remove large HTML content we don't need
json.content nil
json.description nil

if content = recording.recordable.try(:content)
  json.plain_text_content scrubbed_excerpt(excerpt_and_highlight_matches(content), escape: false)
end

if description = recording.recordable.try(:description)
  json.plain_text_description scrubbed_excerpt(excerpt_and_highlight_matches(description), escape: false)
end
```

The type-specific partial emits `content`/`description`, then both are **unconditionally overwritten with `nil`** — deliberately, to keep large HTML out of the search payload. The replacements are emitted under different names.

SDK side (`spec/basecamp.smithy`, `structure SearchResult`):

```
content: String
description: String
```

`plain_text_content` / `plain_text_description`: **zero occurrences** in `spec/basecamp.smithy`.

## Modeling caveat — the name is misleading

`plain_text_content` is **not plain text**. Tracing `excerpt_and_highlight_matches` (`app/helpers/searches_helper.rb:49`) and `scrubbed_excerpt` (`app/helpers/content_helper.rb:101`), the value is:

1. `to_plain_text` of the rich text (or `HtmlScrubber.decode(...).squish` for non-ActionText sources),
2. `html_escape_once`-d,
3. query matches wrapped in **`<mark class="circled-text"><span></span>…</mark>`** (`mark_hits_in`, `searches_helper.rb:142`) — skipped for stop-word-only matches,
4. truncated to a `SNIPPET_SIZE = 300` window chosen to contain the most highlightable hits,
5. passed through `scrubbed_excerpt(..., length: 300, escape: false)`.

So it is a **search-highlighted, HTML-bearing, ~300-char excerpt**, not the full body and not plain text. Whatever we name the SDK field, the doc comment has to say this — a consumer that renders it as literal text will show `<mark>` tags, and one that assumes full content will silently truncate at 300 chars.

Also note both are **conditional** (`if content = recording.recordable.try(:content)`), so they are optional, not `@required`.

## Proposed shape

```
structure SearchResult {
  ...
  /// Always `null` on search results — BC3 strips the full rich-text body from
  /// the search projection. Use `plain_text_content`.
  content: String
  /// Always `null` on search results. Use `plain_text_description`.
  description: String

  /// Search-highlighted excerpt of the recording's `content` ... (~300 chars,
  /// `<mark class="circled-text">` around query matches). Present only when the
  /// recordable has a `content` attribute. NOT plain text despite the name.
  plain_text_content: String
  /// See `plain_text_content` — the description-attribute counterpart.
  plain_text_description: String
}
```

Open questions for whoever picks this up:

1. **Do we `@deprecated` `content`/`description` on `SearchResult`?** They are permanently null in this projection. Deprecating them gives the per-language deprecation matrix (#406) a real signal instead of silently handing back null. Weigh against them being live and meaningful on every *other* recording shape — the deprecation is projection-specific, and our current mechanism is per-member, so this should be fine, but it wants a look.
2. **Naming.** Keep BC3's `plain_text_content` wire name (consistency, greppability) even though it is inaccurate? I lean yes — wire fidelity beats editorializing, and the doc comment carries the correction.
3. Additive and read-only, so no wire change and no breaking-change label; regeneration across six SDKs plus a fixture/decoding test.

## Absorption checklist

- [ ] Model `plain_text_content` / `plain_text_description` on `SearchResult` (optional, non-nullable)
- [ ] Document the highlight markup, the 300-char cap, and the always-null `content`/`description`
- [ ] Decide the `@deprecated` question on `content`/`description`
- [ ] Regenerate all six SDKs; decoding coverage incl. a highlighted-excerpt fixture
- [ ] Add a `spec/api-gaps/` entry recording the BC3 contract, or fold into the search entry

## References

- bc3 `c308693171`: `app/views/api/searches/show.json.jbuilder`, `app/helpers/searches_helper.rb`, `app/helpers/content_helper.rb`
- #464 (the #428 correction that surfaced this)
- `spec/api-gaps/search-filter-additions.md` (`absorbed-in-sdk`; covers request params, not this response gap)

jeremy:
Closes #465. **Stacked on #486** → #484.

## The contract

`app/views/api/searches/show.json.jbuilder` renders the recording's own partial, then:

```ruby
# Remove large HTML content we don't need
json.content nil
json.description nil

if content = recording.recordable.try(:content)
  json.plain_text_content scrubbed_excerpt(excerpt_and_highlight_matches(content), escape: false)
end
if description = recording.recordable.try(:description)
  json.plain_text_description scrubbed_excerpt(excerpt_and_highlight_matches(description), escape: false)
end
```

Two different shapes, and the difference matters:

- `content` / `description` — overwritten **unconditionally**. Keys always present, values always null → **required *and* nullable**. Optional+nullable would accept a payload that omits them, which BC3 never sends.
- `plain_text_content` / `plain_text_description` — emitted **conditionally** → **optional and non-nullable**. A result whose recordable has no such attribute omits the key rather than sending null.

Nullability follows the `SearchType.key` precedent in `spec/smithy-build.json` (`jsonAdd`: `type: ["string","null"]` + `x-go-type: "*string"`).

## The name lies, and the docs say so

`plain_text_content` is not plain text. `excerpt_and_highlight_matches` (`searches_helper.rb:49`) does `to_plain_text` → `html_escape_once` → wraps each query match in `<mark class="circled-text"><span></span>…</mark>` (`mark_hits_in`, skipped for stop-word-only matches) → truncates to `SNIPPET_SIZE = 300`. It is an HTML fragment. Every SDK's doc comment states this.

## This is breaking — #465 said additive

The issue's point 3 claimed *"Additive and read-only, so no wire change and no breaking-change label"*. That holds for the two new fields, not for adding `@required` to two existing ones. [Correction posted on the issue](https://github.com/basecamp/basecamp-sdk/issues/465#issuecomment-5100957196).

| SDK | Before | After |
|---|---|---|
| Go | `Content string \`json:"content,omitempty"\`` | `Content *string \`json:"content"\`` |
| Python | `content: NotRequired[str]` | `content: str \| None` |
| TypeScript | `content?: string` | `content: string \| null` |
| Ruby | optional attr | in `required_fields` |
| Swift | `public var content: String?`, defaulted | `public let content: String?` — **moves into the required init position** |
| Kotlin | — | unaffected: no generated `SearchResult`; `search` returns `ListResult<JsonElement>` (pre-existing typing gap) |

## The fixture was false

`spec/fixtures/search/results.json` had result 0 with full HTML `content`, result 1 with a real `description`, result 2 with full HTML `content`, and **no** `plain_text_*` anywhere — the exact shape the jbuilder makes impossible. Anything asserting against it was pinning fiction.

Rewritten so all three results carry `content: null` and `description: null` plus the appropriate highlighted excerpt: `plain_text_content` for the Message and Comment, `plain_text_description` for the Todo, each with real `<mark class="circled-text">` markup.

## Verification

- **`go/pkg/basecamp/search.go`** is a hard gate — `make go-check-wrapper-drift` requires every generated JSON tag on the wrapper and every field assigned in `searchResultFromGenerated`. Wrapper updated with both new fields and the pointer change; gate reports `83 pairs walked, 1119 generated fields verified`.
- **Go tests** now assert `Content == nil` / `Description == nil` and that the excerpts carry `<mark class="circled-text">`, rather than the old full-HTML strings.
- **New `swift/Tests/BasecampTests/SearchResultDecodeTests.swift`** pins the discriminating half of required-vs-optional: a payload omitting `content` (or `description`) **must fail to decode** with `DecodingError.keyNotFound` naming that key — an assertion that would pass vacuously under optional+nullable. Also pins that re-encoding keeps the nulls on the wire and that `plain_text_*` stay absent when not sent. Swift generates exactly the right thing here: `try container.decode(String?.self, forKey:)` (key required, value nullable) versus `decodeIfPresent` for the excerpts.
- `make check` exits **0**.

## Deferred

The `@deprecated` question on `content`/`description`. They are permanently null *in this projection* but live and meaningful on every other recording shape; worth its own look rather than riding along.

---
## Summary by cubic
Models search results so `content` and `description` are required and nullable, and adds optional `plain_text_content`/`plain_text_description` with 300‑char highlighted excerpts. Updates OpenAPI/Smithy, fixtures, tests, and all SDKs; breaking for consumers treating `content`/`description` as optional.

- **Migration**
  - Use `plain_text_content`/`plain_text_description` for snippets; they are HTML fragments with `<mark class="circled-text">` highlights.
  - Types updated:
    - Go: `Content`/`Description` are `*string` (required); read `PlainTextContent`/`PlainTextDescription`.
    - Python/TypeScript: `content`/`description` are required `string | null`; `plain_text_*` are optional.
    - Ruby: `content`/`description` are in `required_fields`; `to_h` preserves explicit nulls.
    - Swift: `content`/`description` are required init params; payloads omitting them fail to decode; `plainText*` remain optional.

- **Bug Fixes**
  - TypeScript and Ruby search stubs now include `content: null`/`description: null`; tests assert highlighted excerpts and correct omission when not applicable.
  - Ruby: stubs add required `url`/`app_url`; new test pins `required_fields` and keeps nulls in `to_h`.
  - Go/Swift/TypeScript tests assert excerpt HTML and required‑and‑nullable behavior.
  - Repo hygiene: dropped a committed `node_modules` symlink; fixed `.gitignore` rules to avoid re-adding it.

<sup>Written for commit fba66ddd691292d4151a2fd199b802734e666e70. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/basecamp/basecamp-sdk/pull/487?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jeremy:
Fixed the stub gap — you were right that green proved nothing here.

The TypeScript and Ruby search stubs omitted `content` and `description` entirely, which under this PR's contract is a payload the API **cannot** produce: BC3 emits `json.content nil` / `json.description nil` unconditionally on every result. Both suites passed anyway, because neither SDK validates required fields at runtime — TypeScript's types are compile-time only and Ruby just reads keys.

Both stubs now carry `content: null` and `description: null`, one result in each carries a `plain_text_*` excerpt, and both assert the **discriminating** property: the keys are *present* and null, not absent — plus that the excerpt appears only for the result type that has the underlying attribute.

**Verified the assertions bite.** Dropping the two keys back out of the TypeScript stub fails with `expected undefined to be null`.

Rebased onto the post-#483 `main`; 18 files.

Issue #477, opened 2026-07-27
ScheduleEntry update omitting participant_ids silently removes all participants


jeremy:
## Summary

`UpdateScheduleEntry` models `participant_ids` as optional, and the public update example omits it. But BC3's update action always calls `replace_participants`. When the key is absent, parameter normalization produces an empty ID list, so an otherwise successful update removes every participant.

This is SDK-reachable silent data loss. It deserves a focused fix independently of the larger Update/Edit/Replace rollout in #374.

## Pinned evidence

Verified against the SDK's current BC3 provenance pin, `c308693171680e8ec9d3ae9e68181ef2ea80f077`:

1. The flat API route exposes `Schedules::EntriesController#update` (`resources :schedule_entries, only: %i[ show update ]`).
2. The action calls `@recording.replace_participants schedule_entry_participants, track: true` unconditionally, outside any `participant_ids` key guard:
   https://github.com/basecamp/bc3/blob/c308693171680e8ec9d3ae9e68181ef2ea80f077/app/controllers/schedules/entries_controller.rb#L50-L56
3. `schedule_entry_participants` falls back from a missing nested `participant_ids` key to `params[:participants].to_s.split(",")`; when both are absent, that is an empty list:
   https://github.com/basecamp/bc3/blob/c308693171680e8ec9d3ae9e68181ef2ea80f077/app/controllers/schedules/entries/base_controller.rb#L34-L37
4. `UpdateScheduleEntryInput.participant_ids` is optional in `spec/basecamp.smithy`.
5. The public update example itself sends `summary`, `starts_at`, `ends_at`, and `all_day` without `participant_ids`:
   https://github.com/basecamp/bc3/blob/c308693171680e8ec9d3ae9e68181ef2ea80f077/doc/api/sections/schedule_entries.md#L323-L348

## Reproduction

1. Create or select a non-recurring schedule entry with at least one participant.
2. Send the documented update shape, omitting `participant_ids`:

```json
{
  "summary": "All Day Meeting",
  "starts_at": "2015-06-04",
  "ends_at": "2015-06-04",
  "all_day": true
}
```

3. The update succeeds, but the entry's participants are replaced with an empty collection.

Every generated SDK can produce this request because the Smithy member is optional.

## Contract decision and fix boundary

Choose and test one explicit contract:

[...]

PR #488, opened 2026-07-28
Absorb bubble_up_url across the four shapes that carry it

jeremy:
**Stacked on #487** → #486 → #484. Prerequisite for the #477 SDK follow-up.

`bubble_up_url` was already at the provenance pin in `my/readings/_reading.json.jbuilder` and modeled on `Reading`. The post-pin drift is the **shared recording and todolist partials** — and requiredness is not uniform across the shapes that carry it.

## Requiredness differs by shape, and the reason is the call site

`app/views/api/recordings/_recording.json.jbuilder:19-21`:

```ruby
if local_assigns[:bubbleupable]
  json.bubble_up_url bucket_recording_bubble_up_url(recording.bucket, recording, api_url_options)
end
```

`app/views/api/todolists/_todolist.json.jbuilder:2`:

```ruby
json.partial! "recordings/recording", recording: recording, commentable: true, boostable: true, bubbleupable: true
```

That is the **only** `bubbleupable` passer in the entire view tree (`grep -rn "bubbleupable" app/views/` returns exactly these two hits), and it passes it **unconditionally**.

| Shape | Requiredness | Why |
|---|---|---|
| **Todolist** | `@required` | every list/show/group path renders the todolist partial, which always passes the flag |
| **TodolistGroup** | `@required` | `todolists/groups/{index,show}.json.jbuilder` render the same partial |
| **Recording** | optional | polymorphic projection; the generic path never passes the flag |
| **SearchResult** | optional | same, with one variant that does — see below |

`SearchResult` is in scope because Todolists **are** searchable: `Search::Type::TYPE_FILTER_MAPPING` maps `"Todo" => %w[Todo Todolist]`, and `Searches::ApiHelper#api_search_result_template_path` falls through to `recording.to_recordable_partial_path`, which for a Todolist resolves to `todolists/todolist` — the partial that passes the flag.

**Adding required response members changes generated public types, so this is breaking.**

## Fixtures — including the field-bearing representatives

Direct shapes: `todolists/{get,list}.json` and `todolist_groups/{get,list}.json`.

The two optional shapes needed representatives that actually carry the field, because their existing fixtures correctly do **not**:

- `recordings/get.json` and both elements of `recordings/list.json` are Messages, which never receive `bubbleupable`. Added a **third, Todolist-shaped element** to `recordings/list.json`.
- Added a **fourth, Todolist search result** to `spec/fixtures/search/results.json` (with the required-and-nullable `content`/`description` and a highlighted `plain_text_description`, per #487).

Both are registered in `spec/fixtures/manifest.yaml` as new targets (`recording-list-todolist`, `search-result-todolist`) and added to `covered_schemas`, so the coverage guard validates them rather than leaving them unchecked. `make check-fixture-coverage` → *22 manifest targets*.

URLs are derived from each fixture's existing `subscription_url`/`url`, so they stay internally consistent and match the real `bucket_recording_bubble_up_url` shape (`.../buckets/<b>/recordings/<id>/bubble_up.json`, from the `recording_actions` concern in `config/routes.rb:231`).

## Go wrappers

`make go-check-wrapper-drift` named all four shapes; all four wrappers now carry `BubbleUpURL` and assign it in their `*FromGenerated` bodies. Required on Todolist/TodolistGroup (no `omitempty`), `omitempty` on Recording/SearchResult. Gate reports `83 pairs walked, 1123 generated fields verified`.

Go tests assert the field **discriminates** rather than merely existing: for every recording and every search result, a `Todolist` must have `BubbleUpURL` set and every other type must not. A regeneration that made it unconditional would fail.

## Smithy examples

`spec/overlays/examples.smithy` is validated for required-member completeness — the first `make generate` failed with `Missing required structure member bubble_up_url` for both `GetTodolistOrGroup` examples. Both updated.

## `spec/api-gaps/recording-bubbleupable-field.md` stays open

It tracks a distinct **boolean** `bubbleupable` field expressing per-user eligibility. `bubble_up_url` is a URL whose presence is decided by the *rendering call site*, so it is constant per shape and cannot answer "can *this* user bubble this up". Annotated the entry to say so, so nobody mistakes this absorption for closing it. `make validate-api-gaps` → clean, 27 entries.

`make check` exits **0**.

---
## Summary by cubic
Adds `bubble_up_url` to `Todolist`, `TodolistGroup`, `Recording`, and `SearchResult` across the API and SDKs. Breaking change: it’s required on `Todolist` and `TodolistGroup`; regenerate SDKs and update fixtures/tests.

- **New Features**
  - Modeled `bubble_up_url` across four shapes; required on `Todolist`/`TodolistGroup`, optional on `Recording`/`SearchResult` (present for Todolist variants). OpenAPI/Smithy updated; Go/Swift/Kotlin/Python/Ruby/TypeScript SDKs regenerated.
  - Fixtures updated and manifest extended with field-bearing reps in `recordings/list.json` and `search/results.json`.

- **Bug Fixes**
  - Made invented Todolist fixtures self-consistent (bucket IDs, Todoset parent, bookmark SGIDs); Smithy examples updated.
  - TypeScript conformance runner: use `performance.now()` and allow a 2ms `TIMER_SLACK_MS` to handle early Node timers.
  - Tests: Ruby now uses canonical `todolists/get.json` and asserts `bubble_up_url`; Python stubs include required `bubble_up_url`; TypeScript adds an explicit assertion.

<sup>Written for commit 8365feabeea134330f693193fbdc97121c24d5aa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/basecamp/basecamp-sdk/pull/488?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jeremy:
Fixed both live findings.

**The fixtures were mine and they were wrong.** I built each Todolist element by cloning a Message element and overwriting only the URLs, so nothing else agreed: URLs said bucket `2085958500` while `bucket.id` said `2085958499`, `parent` was a `Message::Board` rather than a `Todoset`, and the bookmark SGIDs still decoded to the Message's recording id. Every identity-bearing field is now taken from the real `todolists/get.json` — one bucket, one recording id (`1069479519`), a `Todoset` parent, and a bookmark SGID that decodes to the right recording.

**The 1,999 ms failure was a measurement bug, not a flake to re-run.** The TypeScript runner timestamped requests with `Date.now()`, which floors to whole milliseconds; two floored reads bracketing a 2,000 ms sleep can legitimately differ by 1,999. That is why only the TypeScript runner ever saw it — Go uses nanosecond `time.Time`. Switched to `performance.now()`. The assertion is untouched and the SDK really does sleep the full interval; three consecutive clean runs of the runner.

Also rebased onto the restacked #487 (which is now on the post-#483 `main`).

jeremy:
Correcting myself on the 2,000 ms assertion.

I said `Date.now()` truncation was the cause and that three clean runs proved it fixed. That was wrong — the assertion failed again on a later run at **1999.8753749999996 ms**. Sub-millisecond, so not a rounding artifact: libuv rounds a timer's deadline down internally, so `setTimeout(2000)` can genuinely elapse in slightly under 2,000 ms of wall clock. Go's timers don't fire early, which is why only this runner ever saw it.

`performance.now()` is still correct — `Date.now()` truncation was a real second source of noise — but it needed a documented `TIMER_SLACK_MS` of 2 ms on the comparison, which is now in place with the reasoning inline. That allowance cannot hide a real regression: a dropped `Retry-After` halves the backoff (~1,000 ms off) and a missing delay misses by the whole interval.

Three consecutive clean full-runner runs again — but I'm noting the number rather than resting on it, since that's exactly what I did last time.

PR #489, opened 2026-07-28
Cards: give the plain update name to a merge-safe composite

### LINKED ISSUE #467: Cards.update silently erases due_on on every sparse update (BC3 forced-replace)
## Summary

**`Cards.update` silently erases a card's due date.** Any sparse update — the only kind every SDK can send — clears `due_on` as a side effect:

```go
card, _ := client.Cards.Update(ctx, cardID, generated.UpdateCardRequestContent{Title: "New title"})
// card.DueOn is now nil. The caller never mentioned due_on.
```

This is a live data-loss bug against BC3 as pinned, not a theoretical one, and it is currently untracked.

## BC3 evidence (`c308693171`, the SHA `spec/api-provenance.json` pins)

`app/controllers/kanban/cards_controller.rb`:

```ruby
def card_update_params
  # sets due_on to nil if it's not included in the params
  { due_on: nil }.merge(card_params)
end

def card_params
  params.require(:kanban_card).except(:assignees, :assignee_ids).permit(:title, :content, :due_on)
end
```

BC3 documents the behavior in its own comment. `card_params` only carries keys actually present in the request, so any body omitting `due_on` merges over the `nil` default and the date is destroyed.

## The op is *hybrid*, which is why this is not a Todos repeat

`UpdateCard` mixes three different semantics in one endpoint:

| Field | Semantics | Mechanism |
|---|---|---|
| `title`, `content` | **merge** | absent key ⇒ absent from `card_params` ⇒ `recordable.changing(...)` doesn't touch it |
| `assignee_ids` | **merge, key-guarded** | `update_assignees?` tests `params[:kanban_card].has_key?(:assignee_ids)` |
| `due_on` | **forced replace** | `{ due_on: nil }.merge(...)` |

So `UpdateCard` must **not** be renamed to `Replace*` the way `UpdateTodo` → `ReplaceTodo` was in #375. Only `due_on` behaves like a replace.

## Do not copy Todos' full-body resend algorithm

A full-body resend is wrong here. **Correct algorithm: resend only the caller-specified `title` / `content` / `assignee_ids`, plus the fetched `due_on` when the caller did not address it.**

The reason is *not* spurious notifications — I checked, and that claim (from an earlier draft of this analysis) is false. `change_assignees` does set `@assignees_changed = true` unconditionally (`recording/assignable.rb:102`) and `after_commit :notify_recordable_of_assignees_change` does fire, but it calls `recordable.assignees_changed(self)`, which is an **empty default** on `Recordable` (`concerns/recordable.rb:107`) that **no model overrides**, `Kanban::Card` included. Nothing notifies today.

The real reasons to keep the body sparse:

1. **A redundant resend can silently unassign someone.** `find_assignees` is `@board.reachable_people(include_admissible_people: true).where(id: assignee_ids)`. If a current assignee has since lost board access, they are excluded from `find_assignees`, so `replace_assignees` computes `removed: assignees - people` **including them**, and `unassign` destroys their assignment. Echoing back assignee IDs the caller never mentioned imports this hazard into an update that had nothing to do with assignments.
2. **Lost-update surface.** Every field the composite echoes is a field where a concurrent writer's change gets clobbered by our stale GET. Sparse bodies keep that window as narrow as the caller's actual intent.
3. **Needless locking.** `change_assignees` wraps in `with_lock`, and `assign`/`unassign` each take `with_lock` again — row locks taken for a no-op. (`assign` does filter to `Array(people) - assignees`, so an unchanged set creates no assignments and no tracked events. The work is wasted, not destructive.)
4. **Latent, not live, on notifications.** Because `@assignees_changed` is set even for a no-op, the moment anyone overrides `assignees_changed` on `Kanban::Card` the no-op resend starts firing it. Not a reason to block, but a reason not to build on the assumption.

### Gate A — the generated `update` already owns the method name

`UpdateCard` generates a **public `update`** in all five generated service layers:

| SDK | Site |
|---|---|
| TypeScript | `typescript/src/generated/services/cards.ts:116` |
| Kotlin | `kotlin/sdk/src/commonMain/kotlin/com/basecamp/sdk/generated/services/Cards.kt:40` |
| Ruby | `ruby/lib/basecamp/generated/services/cards_service.rb:26` |
| Python | `python/src/basecamp/generated/services/cards.py:21` (sync), `:79` (async) |
| Swift | `swift/Sources/Basecamp/Generated/Services/CardsService.swift:52` |

Todos only had a free `update` because the operation was **renamed** to `ReplaceTodo`, so its generated method became `replace`. That escape hatch is closed here — Cards must keep `UpdateCard` because BC3 is hybrid.

**Proposal:** keep the wire identity `UpdateCard`, but **generator-rename its SDK method** to something accurate — `put` or `updateVerbatim` — across all five generators. The hand-written composite then owns `update` and internally does GET → generated PUT. **Do not name it `replace`.**

Swift makes this non-optional: it cannot redeclare the same signature, and an overload would leave the unsafe method public.

### Gate B — intentional-clear is currently *unrepresentable*, and fixing that is itself breaking

This is harder than "pick a representation". Today:

```go
type UpdateCardRequestContent struct {
	DueOn types.Date `json:"due_on,omitempty"`   // value type + omitempty
	...
}
```

A zero `types.Date` is **omitted**, so today the only way to express "no due date" is omission — which currently works only *because* the server treats omission as clear. The moment the composite flips omission to mean "preserve", **clearing becomes unrepresentable** in Go without changing the field's type. Same presence problem affects every empty scalar in the body.

So Gate B requires deciding the optional/pointer/sentinel representation **and** accepting that the request-type change is a second breaking change layered on the behavior change. Settle this before writing the composite.

## Required tests

- existing date preserved across an unrelated update
- existing null stays null
- explicit clear
- explicit new date
- omitted assignees **absent** from the PUT body
- explicit empty assignees **present** and clearing

## Composite declaration and wiring

A composite must be *declared*, not just implemented:

- `SPEC.md` §18 "Current composites" (line 1244 today names only Todos `update`/`edit`) and the AGENTS architecture table
- TS import/export + client registration
- Ruby Zeitwerk prepend
- Python **sync and async** client wiring
- Kotlin `EXTENSIBLE_SERVICES` + `HAND_WRITTEN_SERVICES`
- Swift same-module extension

Hook identity across the GET+PUT pair, and the **non-atomicity** of the composite, must be covered and documented.

**Conformance:** `cards_write.json` asserting `requestCount: 2`, GET@0 / PUT@1, present `due_on`. The vocabulary already exists from #375. Dispatch in five runners; Swift mirrors natively per SPEC §18 condition 4. The case must **fail against the pre-fix wrapper**.

## Relationship to #374

#374's triad rollout covers the genuine *replace* recordables. Cards is hybrid and does **not** get a `Replace*` rename — it is tracked here instead. See #374 for the revised per-recordable contract table.

jeremy:
Closes #467. **Stacked on #488** → #487 → `main`.

BC3 builds a card's update params as `{ due_on: nil }.merge(card_params)` (`kanban/cards_controller.rb`), so **any** update whose body omits `due_on` erases the card's due date. A sparse PUT — the natural thing to write, and what every generated SDK produced — silently destroyed data.

## The shape

The plain name goes to the safe thing:

| | `update` (composite) | `updateVerbatim` (raw) |
|---|---|---|
| `dueOn` **unaddressed** | GET, then resend the fetched date | BC3 clears it — the bug, honestly exposed |
| `dueOn` **explicitly empty/null** | clear (one PUT, no GET) | n/a |
| `dueOn` **a date** | set (one PUT, no GET) | set |

The extra GET is paid for **only** when the caller left `dueOn` alone — the one case where the API would otherwise destroy something.

The composite deliberately does **not** resend everything. BC3 filters incoming assignee IDs through `reachable_people`, so echoing assignees back would silently unassign anyone who has since lost board access. Only the caller's own `title`/`content`/`assignee_ids` go out, plus `due_on`.

## Clear is encoded by omission, not by `null`

The plan called for Go to send `{"due_on": null}` with a new SPEC exception. **That doesn't work**, and the recon that found it changed the design:

- `SPEC.md` §18 forbids it outright.
- Five of six SDKs strip nulls *structurally* before the wire — Python `_compact`, Ruby `compact_params`, Kotlin `?.let`, TypeScript's `JSON.stringify` dropping `undefined`, Swift `encodeIfPresent`. One shared conformance fixture could not assert null-for-Go and absent-for-everyone-else.
- `due_on: ""` was the fallback, but `SPEC.md` §5 already records that for the Todos sibling `""` is a **date-format error**.

Omission is correct *and* universally expressible: BC3 nils an omitted due date, which is exactly what a caller asking to clear wants. So no SPEC exception was needed — §18 instead gained a note that body compaction is **not** relaxed for composites. `""`/`null` stays the caller-facing sentinel; omission is the wire encoding.

## Naming, and why the TS type moved

Renaming the generated TS method also renames its request interface, because the TS generator derives interface names from the *method* name: `UpdateCardRequest` → `UpdateVerbatimCardRequest`. Rather than fight that, the composite claims the plain name — and it earns it, because the two types genuinely differ: the composite's `dueOn` is `string | null` (it can express "clear"), the verbatim one's cannot. So `UpdateCardRequest` keeps working for existing callers and is strictly **wider** than before; `UpdateVerbatimCardRequest` is additive.

Kotlin (`UpdateCardBody`), Swift (schema-derived `UpdateCardRequest`), Python and Ruby (kwargs) are unaffected.

Presence detection is language-native: Go `*string`, TS `string | null`, Ruby/Python `nil`/`None` with `""` to clear, Kotlin nullable with `""` to clear, and Swift a **`DueDate` enum** (`.preserve` / `.clear` / `.on`) — an optional cannot carry three states, and `.preserve` being the default means the safe thing is what you get by saying nothing.

## Conformance — two cases, distinct dispatch identities

`conformance/tests/cards_write.json`, run by **all five** executable runners:

- `UpdateCard` → `requestCount: 2`, GET@0, PUT@1, `due_on` resent, `assignee_ids`/`content` absent.
- `UpdateCardVerbatim` → exactly one PUT, no GET, `due_on` absent.

**The second case is the load-bearing one.** Without it, later generator drift could turn both public methods into composite behaviour and nothing would notice. Swift isn't in `make conformance`, so `CardsServiceExtensionsTests` mirrors both plus the clear/set paths.

## Deviation from the plan, stated plainly

The plan asked for a field-specific enhancer pass pointerizing all three `UpdateCardRequestContent` members. **I skipped it**, because it buys no wire behaviour: Go's composite goes through the `map[string]any` + `UpdateCardWithBodyWithResponse` path (sanctioned by §18 rule 1), which the typed generated body never touches. Pointerizing would also not have fixed the thing it was aimed at — `AssigneeIds []int64` keeps `omitempty` regardless, so an empty non-nil slice would still be dropped, breaking the already-shipped "clear assignees" contract. If the intent was SPEC §10 static-typing compliance rather than behaviour, that's worth its own issue.

## Tests

Go: two existing tests moved to `UpdateVerbatim` (they encoded verbatim semantics); six new composite tests — preservation, explicit clear, explicit date, explicit empty content, explicit empty assignees. TypeScript and Ruby unit tests likewise split across both paths. The TS one had been failing only because MSW had no GET handler — which is precisely the read-before-write being added.

`make check` exits **0** (verified exit code).

---
## Summary by cubic
Give the plain `cards.update` name to a merge-safe composite across all SDKs. It preserves `due_on` when you omit it and keeps the raw single-PUT path as `updateVerbatim` (fixes #467).

- **New Features**
  - `update` now GETs the card and resends `due_on` only when you didn’t set it; it does not echo `assignee_ids` or `content` — only your fields plus `due_on` are sent.
  - The raw single-PUT method is renamed to `updateVerbatim` in all SDKs; the composite owns the plain `update` name.
  - TypeScript wires `CardsService` through `services/cards-extensions`, exports `UpdateVerbatimCardRequest`, and keeps `UpdateCardRequest` for the composite.
  - Swift adds a `DueDate` enum (`.preserve`/`.clear`/`.on`); Go’s `UpdateCardRequest` scalars are presence-bearing pointers; Ruby/Python/Kotlin expose `update` as the composite and keep `update_verbatim`/`updateVerbatim` for the raw path.
  - Conformance adds `UpdateCard` (composite) and `UpdateCardVerbatim` (raw) cases; fixtures cover preserve/clear/empty; SPEC documents the merge-safe surface and that body compaction stays strict for composites.

- **Migration**
  - Existing `update` calls now do a read-before-write. Use `updateVerbatim` to keep the old single-request behavior.
  - To clear the due date: TypeScript `dueOn: null`; Go `DueOn: ptr("")`; Ruby/Python `due_on: ""`; Kotlin `dueOn = ""`; Swift `.clear`.

<sup>Written for commit c21e6a87883387202912450a93d6a3ce6617e35b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/basecamp/basecamp-sdk/pull/489?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -9,10 +9,10 @@
 | **TypeScript SDK** | Production-ready | 46 generated services, openapi-fetch based |
 | **Ruby SDK** | Production-ready | 46 generated services |
 | **Swift SDK** | Production-ready | 46 generated services, URLSession-based |
-| **Kotlin SDK** | Production-ready | 46 accessors over 46 generated service classes (`todos` exposes a composite subclass of the generated `TodosService`), Ktor/KMP-based |
+| **Kotlin SDK** | Production-ready | 46 accessors over 46 generated service classes (`todos` and `cards` expose composite subclasses of their generated services), Ktor/KMP-based |
 | **Python SDK** | Production-ready | 46 generated services, httpx-based |
 
-All six SDKs share the same architecture: **Smithy spec -> OpenAPI -> Generated services**. All wire operations are generated. The only hand-written runtime API methods are sanctioned composites that call generated wire methods exclusively (today: the merge-safe Todos `update`/`edit`) — see SPEC.md §18 "Hand-Written Composite Methods" for the rules they must satisfy.
+All six SDKs share the same architecture: **Smithy spec -> OpenAPI -> Generated services**. All wire operations are generated. The only hand-written runtime API methods are sanctioned composites that call generated wire methods exclusively (today: the merge-safe Todos `update`/`edit` and Cards `update`) — see SPEC.md §18 "Hand-Written Composite Methods" for the rules they must satisfy.
 
 ---
 
@@ -38,6 +38,7 @@ All 226 operations across the ~46-service per-SDK layer are generated. Hand-writ
 | HTTP helpers, pagination, hooks | `src/services/base.ts` | `lib/basecamp/services/base_service.rb` | `Sources/Basecamp/Services/BaseService.swift` | `sdk/.../services/BaseService.kt` | `src/basecamp/generated/services/_base.py` |
 | OAuth flows (not in OpenAPI spec) | `src/services/authorization.ts` | `lib/basecamp/services/authorization_service.rb` | — | `sdk/.../oauth/*.kt` | `src/basecamp/services/authorization.py` |
 | Merge-safe Todos composites (update/edit over generated get+replace; SPEC.md §18) | `src/services/todos-extensions.ts` | `lib/basecamp/services/todos_extensions.rb` | `Sources/Basecamp/TodosServiceExtensions.swift` | `sdk/.../services/TodosService.kt` | `src/basecamp/services/todos.py` |
+| Merge-safe Cards composite (update over generated get+updateVerbatim; SPEC.md §18) | `src/services/cards-extensions.ts` | `lib/basecamp/services/cards_extensions.rb` | `Sources/Basecamp/CardsServiceExtensions.swift` | `sdk/.../services/CardsService.kt` | `src/basecamp/services/cards.py` |
 
 Hand-written service files in `src/services/` (TS) and `lib/basecamp/services/` (Ruby) beyond the tables above are NOT loaded at runtime. They exist only as reference implementations.
 
--- a/SPEC.md
+++ b/SPEC.md
@@ -275,6 +275,32 @@ account, attachments, automation, boosts, campfires, cardColumns, cardSteps, car
 
 The OpenAPI spec uses 12 coarse tags (e.g., `Automation`, `Todos`, `Files`). The service generators split these into 46 fine-grained services using a two-table mapping: `TAG_TO_SERVICE` (tag → default service name) and `SERVICE_SPLITS` (tag → {service → [operationIds]}). For example, the `Todos` tag splits into `Todos`, `Todolists`, `Todosets`, `TodolistGroups`; the `Files` tag splits into `Attachments`, `Uploads`, `Vaults`, `Documents`. These mappings are defined in each language's generator script and produce identical service sets across SDKs.
 
+### Merge-Safe Write Surface (Cards)
+
+BC3 builds a card's update params as `{ due_on: nil }.merge(card_params)`
+(`kanban/cards_controller.rb`), so **any** update whose body omits `due_on` erases the card's due
+date. A sparse PUT — the natural thing to write, and what every generated SDK produced — is
+therefore destructive.
+
+- **`update`** — merge-safe. GETs the card and resends the existing `due_on` when the caller left it
+  unaddressed, then PUTs. The extra GET is paid for only in that case; naming the due date
+  explicitly skips it. `due_on` is tri-state: unaddressed preserves, an explicit empty clears, a
+  date sets. Clearing is encoded by **omitting** `due_on` — never by sending null (§18).
+- **`updateVerbatim`** — the raw single PUT, no read-before-write. Sharp by construction: omitting
+  `due_on` clears it.
+
+The composite deliberately does **not** resend everything. BC3 filters incoming assignee IDs through
+`reachable_people`, so echoing assignees back would silently unassign anyone who has since lost board
+access; only the caller's own `title`/`content`/`assignee_ids` go out, plus `due_on`.
+
+Not atomic: a concurrent due-date change landing between the GET and the PUT is overwritten with the
+value the call read. The window is one round-trip.
+
+Presence detection is language-native: Go `*string` (nil preserves, pointer-to-empty clears),
+TypeScript `dueOn?: string | null`, Ruby/Python `nil`/`None` kwarg defaults with `""` to clear,
+Kotlin nullable parameters with `""` to clear, Swift a `DueDate` enum (`.preserve`/`.clear`/`.on`)
+because an optional cannot carry three states.
+
 ### Merge-Safe Write Surface (Todos)
 
 The `PUT /{accountId}/todos/{todoId}` endpoint is **full replace, omission clears** (spec operation `ReplaceTodo`, `content` required, declared via `x-basecamp-write-semantics: {mode: "replace", clearsOmitted: true}` and the `write` clause in `behavior-model.json`). Every SDK exposes a three-method, two-state surface over it:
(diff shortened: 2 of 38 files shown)


jeremy:
Addressed the proof-matrix gap, and it turned up the runner defect you predicted.

## The matrix is now complete

`cards_write.json` had only preservation and raw-omission. Added three cases, all five runners green:

| case | asserts |
|---|---|
| `update-explicit-clear` | 1 request, no GET, `due_on` **absent** |
| `update-explicit-empty-content` | `content` **present and `""`** |
| `update-explicit-empty-assignees` | `assignee_ids` **present and `[]`** |

## The Go runner really did collapse empty to absent

As you said, adding the empty-content case exposed it. The runner built the request with

```go
if v := getStringParam(tc.RequestBody, "content"); v != "" {
```

which is precisely the distinction the pointer fields exist to preserve — an explicit clear was indistinguishable from an omission. Replaced with a presence-based `cardUpdateRequest` helper shared by both dispatch arms.

**Verified the new case bites.** Reintroducing the non-emptiness test:

```
FAIL: update-explicit-empty-content: an explicit empty content is sent, not dropped
      Expected request body field "content" on request index 0, but it was absent
```

## Python and Kotlin now have direct service tests

Both previously relied on the shared fixture alone. `python/tests/services/test_cards.py` (8 tests, sync **and** async) and `kotlin/.../CardsServiceTest.kt` (5 tests) cover preservation, explicit clear, explicit date, explicit empty content/assignees, and verbatim's single PUT.

## `comment_count`

Dropped from the write fixture's mock bodies. It is worth being precise about what it is: **not** purely fictional — it is absent from the `Card` schema (which has `comments_count`) but *is* modeled on the hand-written Go wrapper (`go/pkg/basecamp/cards.go:51,105`). So it is a wrapper/spec divergence. It has no place in a write-semantics fixture either way; the three source card fixtures still carry it and I've left those alone as a separate finding rather than fold an unrelated cleanup in here.

## The plan now agrees with the code

You were right that leaving the closeout and the controlling plan disagreeing is not acceptable. `handoff-…-cryptic-scott.md` §Phase 5 is amended rather than silently diverged from: the enhancer pass is marked **DESCOPED** with the reason (nothing constructs the typed body, and it would not have fixed `AssigneeIds` anyway, which keeps `omitempty` regardless and would have broken the shipped clear-assignees contract), and items 2–4 are marked **REVISED** for the omission-not-null encoding.

`make check` exits **0**. Rebased onto the restacked #488.

PR #491, opened 2026-07-28
Repin provenance to current bc3 HEAD; pin the participant_ids contract

### LINKED ISSUE #477: ScheduleEntry update omitting participant_ids silently removes all participants

jeremy:
Closes #477. **Stacked on #489** → #488 → #487 → `main`.

basecamp/bc3#12425 merged as `dffa7e11b3`, fixing the schedule-entry participant wipe server-side. This is the SDK half.

## The repin follows the sync-time rule, not the fix commit

Per repository policy the pin records the **upstream HEAD the spec was synced against**, not the SHA that happened to contain the fix:

1. Fetched `bc3 origin/master` → **`dffa7e11b3337b17454d9a82301be3e94226a858`** (2026-07-28).
2. Verified the #12425 merge is an **ancestor** — `git merge-base --is-ancestor` passes. (Here they coincide, since the merge is currently HEAD; the check matters because that will stop being true the moment anything else lands.)
3. Triaged the drift (below).
4. Pinned `spec/api-provenance.json` to that HEAD + date; `make provenance-sync` propagated it. `API_VERSION` tracked `2026-07-26` → `2026-07-28`.
5. The #12425 merge SHA is recorded here as **fix evidence**, not as the pin value.

## Drift triage — 116 commits, six touch the API

`git log c308693171..dffa7e11b3 -- app/views/api config/routes.rb doc/api`:

| commit | disposition |
|---|---|
| `dffa7e11b3` the #12425 fix | absorbed here |
| `012ae895d1`, `0e4fe32dcd`, `21b3718af9` bubble_up_url | absorbed by #488 |
| `e4bf93c3c2` + `f697d8604d` | **registered** — see below |

### The one genuinely new gap

`e4bf93c3c2` added `everything/assignments`; `f697d8604d` then renamed it to `everything/tasks` before it ever appeared under a pin. Net: a **new** Everything aggregate at

```
GET /:account_id/tasks.json
GET /:account_id/tasks/results.json
```

that the SDK does not model. Registered as `spec/api-gaps/everything-tasks-aggregate.md` (`partial-coverage`) rather than carried silently across the repin — absorbing two new list endpoints is Phase-4-sized work, not repin work.

**It is not a break of `my/assignments`,** and the entry says so explicitly. "Move /assignments to /tasks" reads alarming against an SDK shipping `GetMyAssignments`, but the moved routes are the ones inside `scope module: :everything`. The SDK's assignment operations use `/:account_id/my/assignments.json`, untouched across the entire drift range — `git diff c308693171..dffa7e11b3 -- config/routes.rb` contains only the two `tasks` additions.

## Semantics, documented where they're consumed

`UpdateScheduleEntryInput.participant_ids` now carries the contract — **omission preserves, an empty array clears** — along with why that guarantee is only days old. Only the *update* input gets it; on create there is nothing to preserve.

## Conformance

New `conformance/tests/schedule_entries_write.json`, dispatched by **all five** executable runners plus a Swift mirror:

- **omitted `participant_ids`** → one PUT, `requestBodyAbsent`. This is the load-bearing one: BC3 preserves participants *only because the key is missing*, so a compactor emitting `null`, or a default emitting `[]`, would clear the list on the server.
- **explicit `participant_ids: []`** → one PUT with the empty array **present**. Ruby's `compact_params` and Python's `_compact` strip nil/None but not empty collections, which is what keeps "not addressed" distinct from "addressed and cleared".

Kotlin initially **SKIPped** both, because my mock body lacked required `ScheduleEntry` members (`parent`, `bucket`, `creator`, `description_attachments`) — the same impossible-payload defect being fixed elsewhere in this stack, and a good argument for Kotlin's strict decoding. Fixed rather than accepted as a skip; a skip is not a pass.

## Division of proof

Conformance proves **SDK wire presence** only. The server-side guarantee — that omission preserves and `[]` clears — is proven by the BC3 request tests in #12425, where the preserve test fails against the unpatched controller with exactly the wipe signature (`Expected: [525460207, 950737456], Actual: []`).

Full `make check` green.

---
## Summary by cubic
Repinned API provenance to BC3 HEAD (2026-07-28) and documented the `UpdateScheduleEntry` `participant_ids` contract: omission preserves participants, `[]` clears. Marked Everything “tasks” as web-only and added conformance plus SDK guards to only send `participant_ids` when present.

- **Bug Fixes**
  - Schedule entry updates: `participant_ids` is presence-bearing. SDKs now omit the key unless provided; an explicit `[]` is sent to clear.
  - Conformance: added `conformance/tests/schedule_entries_write.json` and wired into Go, Python, Ruby, TypeScript, Kotlin; added a Swift mirror test.

- **Refactors**
  - Repinned provenance to BC3 `dffa7e11b3` and bumped API version to `2026-07-28` across Go, Kotlin, Python, Ruby, Swift, and TypeScript.
  - Recorded Everything “tasks” as a web view, not an API resource (`spec/api-gaps/everything-tasks-aggregate.md`) to prevent re-flagging.

<sup>Written for commit d54f4141ba15c0441d51441bca9828e16779cc4d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/basecamp/basecamp-sdk/pull/491?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jeremy:
You're right that the `/tasks` gap was fiction, and the reclassification is in.

At the pinned `dffa7e11b3`, `Everything::AssignmentsController#index` is an empty action and the results controller declares `layout "turbo_rails/frame"` — it exists to serve a Turbo Frame for the filter UI. Every view under `app/views/everything/assignments/` is `.html.erb`; no `respond_to`, no JSON branch, no jbuilder, nothing in `doc/api/`. A Rails route tolerating a `.json` suffix does not create a JSON representation.

Now `status: confirmed-not-api-resource`, and the **provisional response shape and SDK absorption plan are removed** rather than left as a sketch of a payload that does not exist. The entry's purpose is inverted: it records a *negative* result so these routes don't get re-flagged at the next sync, since sitting inside `scope module: :everything` makes them look like siblings of the modeled JSON aggregates. The "not a break of `my/assignments`" note is kept, which was the useful part.

Noted on `b238a074` being post-sync drift — it's outside the `dffa7e11` run-boundary pin and belongs to the next sync, so I've left it alone.

Rebased onto `main` (#487/#488/#489 all merged) and retargeted.

Issue #496, opened 2026-07-29
TS conformance runner: delayBetweenRequests flakes 1ms under minDelay


jeremy:
The `delayBetweenRequests` assertion in `conformance/runner/typescript/runner.test.ts` (~line 705) measures `Date.now()` deltas between tracked requests and asserts `delay >= minDelay` exactly. Node timers can fire ~1ms early relative to `Date.now()` rounding (timers schedule on the monotonic clock at ms granularity), so a 2000ms Retry-After wait occasionally measures 1999ms.

Observed on PR #376 CI (run 30435999224): `[GET operation retries on 429 with Retry-After] expected delay >= 2000ms, got 1999ms` — passed on re-run and on the superset branch at the same code.

Fix suggestion: tolerate sub-timer-granularity earliness, e.g. `toBeGreaterThanOrEqual(minDelay - 5)` with a comment explaining the Date.now()/timer-clock rounding mismatch, or measure with `performance.now()` on both sides. Other runners' delay assertions are worth auditing for the same exactness.

jeremy:
Both halves are now on main: 058fd1172 (#488) fixed the reported TypeScript flake with the TIMER_SLACK_MS tolerance, and 78f8e941d (#530) closed the same flake class in the Python and Ruby runners by moving delay measurement to monotonic clocks (time.monotonic / Process.clock_gettime(CLOCK_MONOTONIC)), matching Go's already-safe time.Time subtraction. Wall clocks remain only where absolute timestamps are reported.

PR #530, opened 2026-07-31
Measure retry delays with monotonic clocks in the Python and Ruby conformance runners

jeremy:
## The flake class

`delayBetweenRequests` assertions compare captured request timestamps against an exact minimum-delay bound. When the capture uses a **wall clock**, any NTP step or slew between two captures can make a real sleep of `min_delay` read marginally short, failing the assertion — a nondeterministic flake unrelated to SDK behavior.

#496 reported exactly this in the TypeScript runner; 058fd1172 (#488) patched it there with a `TIMER_SLACK_MS = 2` tolerance (a libuv early-fire workaround). The same class remained latent in two other runners:

- **Python** (`conformance/runner/python/runner.py`): captured `time.time()`, asserted `d < min_delay`
- **Ruby** (`conformance/runner/ruby/runner.rb`): captured `Time.now`, same exact bound

## The fix: Go's approach, not TS's

The Go runner never had this problem — `main.go` subtracts `time.Time` values, which is monotonic in Go. This PR applies the same principle rather than copying the slack-constant hack:

- Python captures `time.monotonic()`
- Ruby captures `Process.clock_gettime(Process::CLOCK_MONOTONIC)`

Both SDKs' retry sleeps wait against the monotonic clock, so measuring on the same clock makes the elapsed delta a true lower bound on the sleep — no tolerance constant needed, and the assertion stays exact.

## Consumer audit

The captured timestamp's **only consumer** in both runners is the delay-delta math (`delays_between_requests`). Nothing exports or reports it as an absolute time — other assertions read `method`/`url`/`headers`/`body`, and reports carry name/pass/fail only. So the field swaps wholesale, renamed `time` → `monotonic_time` so it can't be mistaken for a wall-clock reading. Ruby's now-unused `require "time"` is dropped.

Replay runners capture no timestamps; Kotlin/Swift runners don't implement `delayBetweenRequests`. Go and TS are untouched.

## Evidence

Both runners green with the change, exit codes pinned:

- `make conformance-python` → `REAL_EXIT=0`, 0 failed
- `make conformance-ruby` → `REAL_EXIT=0`, 0 failed

The delay-assertion fixtures (`GET operation retries on 503`, `GET operation retries on 429 with Retry-After`) PASS through the monotonic path in both; the two DownloadURL delay fixtures remain pre-existing intentional skips in both runners.

Closes the Python/Ruby residual of #496 (TS half already fixed by 058fd1172).

---
## Summary by cubic
Switch Python and Ruby conformance runners to monotonic clocks for measuring retry delays to eliminate flaky failures from wall-clock adjustments. Keeps exact minimum-delay assertions without adding tolerance.

- **Bug Fixes**
  - Python: replace `time.time()` with `time.monotonic()` and compute delays from `monotonic_time`.
  - Ruby: replace `Time.now` with `Process.clock_gettime(Process::CLOCK_MONOTONIC)` and compute delays from `monotonic_time`.
  - Rename captured field `time` → `monotonic_time`; remove unused `require "time"` in Ruby.

<sup>Written for commit 22b1241450982b9fe206a9f1bc3f7587393a40cf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/basecamp/basecamp-sdk/pull/530?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/conformance/runner/python/runner.py
+++ b/conformance/runner/python/runner.py
@@ -37,8 +37,8 @@ _MISSING = object()
 class TestTracker:
     requests: list[dict] = field(default_factory=list)
 
-    def record_request(self, *, time: float, method: str, url: str, headers: dict, body: Any = None) -> None:
-        self.requests.append({"time": time, "method": method, "url": url, "headers": headers, "body": body})
+    def record_request(self, *, monotonic_time: float, method: str, url: str, headers: dict, body: Any = None) -> None:
+        self.requests.append({"monotonic_time": monotonic_time, "method": method, "url": url, "headers": headers, "body": body})
 
     def reset(self) -> None:
         self.requests.clear()
@@ -49,10 +49,14 @@ class TestTracker:
 
     @property
     def delays_between_requests(self) -> list[int]:
+        # Elapsed ms between consecutive requests, from monotonic captures.
+        # Wall-clock (time.time) can step or slew mid-test and read a sleep
+        # as shorter than it was — the delay-flake class from #496. Monotonic
+        # deltas mirror the Go runner's time.Time subtraction.
         if len(self.requests) < 2:
             return []
         return [
-            int((b["time"] - a["time"]) * 1000)
+            int((b["monotonic_time"] - a["monotonic_time"]) * 1000)
             for a, b in zip(self.requests, self.requests[1:])
         ]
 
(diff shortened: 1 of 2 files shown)


jeremy:
Bot-review status: Copilot errored on its review attempt and a re-request could not be resolved via API; Codex is at its usage limit today. Zero review threads. Merging on green CI per the standing flaky-reviewer convention.

PR #536, opened 2026-07-31
docs(spec): clarify TypeScript integer waiver

### LINKED ISSUE #533: TS waiver 1B.6: "no clean fix" for >2^53 integer precision is stale against the Node >=22.12.0 floor
## Why

Waiver 1B.6 (`rubric-audit.json:95-100`, as of `dc1b3c590`) reads:

> "JavaScript Number loses precision on integers > 2^53; no clean fix without breaking SDK API surface"

The "no clean fix" half of that note is now stale. #469 raised the TypeScript engine floor to Node `>=22.12.0` (`typescript/package.json:49-51`), and on that floor `JSON.parse` reviver context (`(key, value, context) => ...` with `context.source`) is available — lossless 64-bit decode is **technically feasible**: the reviver sees the raw source digits and can produce a `BigInt` with no precision loss. Verified on Node v22.23.1: `context.source` for `9007199254740993` yields the exact digits, and `BigInt(context.source)` round-trips losslessly.

The waiver's *conclusion* still stands on its other half: emitting `bigint` (everywhere, or past 2^53) is a breaking change to the public `number`-typed ID surface. But the note now asserts something false ("no clean fix"), and the next reader has to re-derive whether the modern option was considered.

## Ask (decide and document — not necessarily code)

Rule the Node >=22.12.0 option explicitly in or out, and update the 1B.6 note to record the decision:

- **Out (keep waiver):** rewrite the note to say lossless decode is feasible on the current engine floor via `JSON.parse` source access, and the waiver is retained solely because `number` → `bigint` is a breaking API surface change (not because no mechanism exists).
- **In (plan the fix):** scope what a lossless surface would look like (opt-in bigint mode, string-typed IDs, or a major-version break) as its own issue; 1B.6 then becomes a pointer to that plan.

Related text that should stay consistent with whatever the note says: SPEC.md:735 (conflict marker), SPEC.md:1748 (waiver table), SPEC.md:1909 (per-language precision table), and `conformance/tests/integer-precision.json` (test description).

lntutor:
## Summary

- clarify that Node 22 reviver source access makes lossless integer decoding technically feasible
- retain the TypeScript waiver because returning `bigint` would break the public `number`-typed API
- keep the rubric, specification, and conformance test description consistent

## Verification

- `python3 -m json.tool rubric-audit.json`
- `python3 -m json.tool conformance/tests/integer-precision.json`
- verified lossless `context.source` decoding on Node 22.22.0
- `git diff --check`

Release notes: Documentation

Closes #533

---
## Summary by cubic
Clarifies the TypeScript integer precision waiver: Node >=22.12 can access raw digits in a JSON.parse reviver, making lossless bigint decoding feasible, but we retain the waiver because returning bigint would break the number-typed API. Syncs SPEC.md, the integer-precision conformance test description, and the rubric note.

<sup>Written for commit ff13a0986dbd02e225baad8658ca72fcadae0a34. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/basecamp/basecamp-sdk/pull/536?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -732,7 +732,7 @@ Comparison is case-insensitive.
 
 All integer IDs must use at least 64 bits of precision (e.g., Go `int64`, Kotlin `Long`, Swift `Int` on 64-bit platforms). Note: Kotlin `Int` is 32-bit and must not be used for IDs — use `Long`. IDs up to 2^53 + 1 (`9007199254740993`) must survive JSON round-trip without precision loss.
 
-`[CONFLICT: JavaScript Number.MAX_SAFE_INTEGER is 2^53 - 1. The TypeScript SDK has a documented known gap — JSON.parse truncates integers beyond this value. The spec prescribes 64-bit precision; TypeScript implementations must document the limitation. See waiver 1B.6 in rubric-audit.json.]`
+`[CONFLICT: JavaScript Number.MAX_SAFE_INTEGER is 2^53 - 1. On the supported Node >=22.12 floor, JSON.parse reviver source access makes lossless bigint decoding feasible, but returning bigint would break the TypeScript SDK's number-typed API surface. The spec prescribes 64-bit precision; TypeScript implementations must document the retained limitation. See waiver 1B.6 in rubric-audit.json.]`
 
 ### Nullable Numeric Dimensions (rich-text attachment width/height)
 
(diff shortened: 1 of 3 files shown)


jeremy (approved):
Reviewed as maintainer against post-#535 main:

- **Accuracy**: The Node >=22.12 claim matches what we verified empirically in #533 — `JSON.parse` reviver `context.source` yields the raw digits, so lossless `bigint` decode is feasible on the supported engine floor. The rewrite correctly retains waiver 1B.6 on its surviving ground (bigint would break the number-typed API surface) and promises no SDK adoption.
- **Consistency**: All four sites from #533 are covered or verifiably fine — SPEC §10 conflict marker and the per-language precision table are updated here; the §19 waiver-table row (| 1B.6 | §10 | 64-bit integer precision |) is index-only with no stale claim; the §19 TS skip-roster line ("Number is 53-bit") stays factually consistent with the new text.
- **Scope**: `integer-precision.json` change is description-only — operation, method, path, and assertions untouched.

Thanks for threading this precisely through the spec, rubric, and conformance description.

jeremy:
Merged as d3422f589. Thank you for a first contribution that hit exactly the right altitude: you took the "decide and document" ask from #533, made the call the evidence supports (waiver retained on API-surface grounds, not mechanism grounds), and threaded it consistently through the spec, the rubric, and the conformance description without touching any test semantics. That is precisely how we hoped that issue would land. Welcome, and we would be glad to see you back.

PR #623, opened 2026-08-03
Page docs: state the contract #617 established, not the divergence it removed

jeremy:
#617 removed the divergence in `page` semantics. It did not remove the 48 doc
comments that describe it, so the public surface of all six SDKs — IDE
autocomplete, godoc, KDoc, generated docs — still tells consumers the opposite
of the contract v0.13.0 exists to establish.

## Before / after

```diff
- Page number for paginating through results. Defaults to 1. Semantics vary by SDK; see SPEC section 8.
+ Page number for paginating through results. Defaults to 1. A positive value selects exactly that page, not a starting offset; see SPEC section 8.
```

and the one short form, on `GetBubbleUpsInput`:

```diff
- Page number. Defaults to 1. Semantics vary by SDK; see SPEC section 8.
+ Page number. Defaults to 1. A positive value selects exactly that page, not a starting offset; see SPEC section 8.
```

The replacement is SPEC §8's own emphasis — "**A positive `page` selects
exactly that page. It is not a starting offset.**" (SPEC.md:858) — rather than
a third phrasing that would then need reconciling with §8 and with the six
READMEs #617 wrote. The `see SPEC section 8` pointer is kept as-is: it already
lands on correct text, because #617 rewrote that section.

## Occurrence count to zero

Two distinct strings carried the stale clause, not one:

| String | Before | After |
|---|---|---|
| `Page number for paginating through results. Defaults to 1. Semantics vary by SDK; …` | 47 | 0 |
| `Page number. Defaults to 1. Semantics vary by SDK; …` | 1 | 0 |
| **`Semantics vary`, anywhere in the repository** | **spec + generated + `openapi.json`** | **0** |

```
$ rg -c "Semantics vary" spec/basecamp.smithy
(no matches, exit 1)
$ git grep -c "Semantics vary"
(no matches)
```

`SPEC.md` carried no copy of the phrasing outside §8, so nothing there needed
touching — and §8 itself is #617's text and is left alone.

## No wire or signature change

`make generate` churned 98 generated files, which is the point: the comment has
to reach consumers' tooling to be worth fixing. But the churn is textual and
nothing else:

- **Operation count unchanged at 240**, before and after.
- **Every changed line in all 99 files** is one of the two old strings going
  out or its replacement coming in. Filtering the full diff for lines matching
  neither returns zero:
  ```
  $ git diff -U0 main | rg '^[+-]' | rg -v '^(\+\+\+|---)' \
      | rg -vc 'Semantics vary by SDK|selects exactly that page, not a starting offset'
  0
  ```
- `openapi.json` moves only `description` values.
- 598 insertions, 598 deletions — balanced, as a pure text swap must be.

Labelled `documentation`, not `breaking`: no wire surface, no signature, no
behavior moves. #617 already shipped the behavior; this only stops the docs
contradicting it.

## Coverage note

Go, Kotlin, Ruby, Swift and TypeScript all carried the stale text and are all
fixed. **Python carried no copy** — its generator emits no per-parameter
docstrings at all, so `page` is undocumented there rather than mis-documented.
Pre-existing and out of scope for a freeze-week doc fix, but worth naming: the
Python `page` keyword is the one surface this PR cannot reach.

## What was deliberately not changed

`Defaults to 1.` is retained. It is true of the wire parameter, which is what
this spec describes. It is worth knowing that at the SDK layer an absent `page`
means "auto-paginate the whole collection" for trait-carrying operations rather
than "page 1" — but that is not universal (`GetMyNotifications` declares `page`
and carries no pagination trait, per §8), so no single sentence covers all 48
sites, and it is a pre-existing imprecision rather than something #617 created.
Left for a separate change if it is judged worth making.

---
## Summary by cubic
Update `page` parameter docs to reflect the unified contract: a positive `page` selects exactly that page, not a starting offset. Removes stale “Semantics vary by SDK” wording across all SDKs to match SPEC §8.

- **Bug Fixes**
  - Replaced old phrasing (47 long-form + 1 short) across `spec/basecamp.smithy`, `openapi.json`, and generated SDKs; kept the SPEC §8 reference.
  - Regenerated Go, Kotlin, Ruby, Swift, and TypeScript outputs so IDE/tooling show the correct docs; Python unchanged (no per-parameter docstrings).
  - No wire or signature changes; diffs touch `description`/doc comments only; operation count stays 240.

<sup>Written for commit f72f977cd1eaef37f512f77db2744d2f6ceda17d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/basecamp/basecamp-sdk/pull/623?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/go/pkg/generated/client.gen.go
+++ b/go/pkg/generated/client.gen.go
@@ -3568,7 +3568,7 @@ type ListClientApprovalsParams struct {
 	// Direction asc|desc
 	Direction *string `form:"direction,omitempty" json:"direction,omitempty"`
 
-	// Page Page number for paginating through results. Defaults to 1. Semantics vary by SDK; see SPEC section 8.
+	// Page Page number for paginating through results. Defaults to 1. A positive value selects exactly that page, not a starting offset; see SPEC section 8.
 	Page *int32 `form:"page,omitempty" json:"page,omitempty"`
 }
 
@@ -3580,19 +3580,19 @@ type ListClientCorrespondencesParams struct {
 	// Direction asc|desc
 	Direction *string `form:"direction,omitempty" json:"direction,omitempty"`
 
-	// Page Page number for paginating through results. Defaults to 1. Semantics vary by SDK; see SPEC section 8.
+	// Page Page number for paginating through results. Defaults to 1. A positive value selects exactly that page, not a starting offset; see SPEC section 8.
 	Page *int32 `form:"page,omitempty" json:"page,omitempty"`
 }
 
 // ListClientRepliesParams defines parameters for ListClientReplies.
 type ListClientRepliesParams struct {
-	// Page Page number for paginating through results. Defaults to 1. Semantics vary by SDK; see SPEC section 8.
+	// Page Page number for paginating through results. Defaults to 1. A positive value selects exactly that page, not a starting offset; see SPEC section 8.
 	Page *int32 `form:"page,omitempty" json:"page,omitempty"`
 }
 
 // ListCardsParams defines parameters for ListCards.
 type ListCardsParams struct {
-	// Page Page number for paginating through results. Defaults to 1. Semantics vary by SDK; see SPEC section 8.
+	// Page Page number for paginating through results. Defaults to 1. A positive value selects exactly that page, not a starting offset; see SPEC section 8.
 	Page *int32 `form:"page,omitempty" json:"page,omitempty"`
 }
 
@@ -3605,7 +3605,7 @@ type GetEverythingCompletedCardsParams struct {
 	// Due Filter by due date: with, without, or overdue. Unrecognized values are ignored.
 	Due *string `form:"due,omitempty" json:"due,omitempty"`
 
-	// Page Page number for paginating through results. Defaults to 1. Semantics vary by SDK; see SPEC section 8.
+	// Page Page number for paginating through results. Defaults to 1. A positive value selects exactly that page, not a starting offset; see SPEC section 8.
 	Page *int32 `form:"page,omitempty" json:"page,omitempty"`
 }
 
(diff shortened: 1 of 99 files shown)


jeremy:
CI note: **`npm Audit (TypeScript SDK)` fails on `main` too**, at the exact merge base `9ddb5875a` — `GHSA-7p8r-x3mc-p8w7`, fast-uri host confusion, one high-severity transitive advisory. It also failed on `eb4d4897e`. Pre-existing, unrelated to this PR, which touches no dependency.

Local `make check` green at `f72f977cd1eaef37f512f77db2744d2f6ceda17d`, real exit code 0, head SHA identical before and after the run. Independent confirmations of the "no structural movement" claim:

- `bc3 route parity OK: 240 SDK routes vs 369 bc3 routes (137 live-marker-backed)` — the operation count is asserted by a gate, not just by my `jq`.
- `API Compatibility` passed in CI.
- `sync-doc-constants self-test: all cases passed.`

Swift and Ruby ran for real locally (`swift test`, 333 tests in the main suite plus the conformance runner; Ruby `1241 runs, 29706 assertions, 0 failures, 0 errors, 0 skips`). **Kotlin locally was `:basecamp-sdk:jvmTest FROM-CACHE`** — a build-cache restore, not an execution, which is what you would expect when the only delta is KDoc that the compiler discards. So the Kotlin evidence here is CI's `Kotlin Tests` job, which ran 3m24s and passed.

Issue #562, opened 2026-08-01
Go wrapper surface types optional timestamps inconsistently (HillChart.UpdatedAt, SearchResult timestamps)


jeremy:
Split out of #560's review, where changing one field drive-by proved to be the wrong venue.

The hand-written Go wrapper surface is inconsistent about how it types **optional** timestamps:

| Type | Field | Go type |
|---|---|---|
| `Card` | `CompletedAt` | `*time.Time` |
| `TimelineEvent` | `CreatedAt`, `UpdatedAt` | `*time.Time` |
| `EverythingFile` | `CreatedAt`, `UpdatedAt` | `*time.Time` |
| `HillChart` | `UpdatedAt` | `time.Time` |

All four schemas mark the field optional, so `HillChart.UpdatedAt` is the outlier: an absent `updated_at` surfaces as `0001-01-01T00:00:00Z`, indistinguishable from a real (if implausible) value. Its `omitempty` is also inert — `encoding/json` never omits a zero `time.Time`, so the key is always emitted on re-marshal.

**Why this was NOT fixed in #560**, and why it needs its own PR: flipping the field to `*time.Time` is a *silent* breaking change for callers. `hc.UpdatedAt.IsZero()` still compiles — Go auto-dereferences the pointer receiver — and then panics at runtime on nil. Verified with a standalone repro. A change with that failure mode needs to be deliberate, announced in release notes, and ideally landed with the other wrapper-surface breaks rather than smuggled into a generated-types PR.

**Options to decide:**
1. Make `HillChart.UpdatedAt` a `*time.Time` (consistent with the other three; breaking, needs a release note).
2. Leave it value-typed and accept that absent reads as the zero time, documenting it — but then the `omitempty` tag should go, since it never fires.

Worth resolving alongside any future audit of value-typed optional fields on the hand-written surface generally; this issue is the timestamp instance, not necessarily the whole class.

🤖 _Drafted with agent assistance; filed by @jeremy._

jeremy:
**Widening this from `HillChart.UpdatedAt` to the class**, after a second instance surfaced in review.

`SearchResult.created_at`/`updated_at` are optional in the schema (`required` is `app_url, content, description, id, title, type, url`), but the wrapper types them as value `time.Time`, so an absent timestamp decodes to `0001-01-01T00:00:00Z` and re-marshals as a value the API never sent.

Current state of the class:

| Type | Field | Go type | Optional in schema |
|---|---|---|---|
[...]

PR #658, opened 2026-08-04
Go: five wrapper timestamps could not represent absence (#620)

### LINKED ISSUE #620: Five more Go wrapper timestamps cannot represent absence, and no gate can see them
Split out of #615, which fixed five of these and did not fix the other five. Reported, not fixed — see "Why this is not a #615 amendment" below.

## Five more wrapper timestamps cannot represent absence

Same defect class as #562: a hand-written wrapper field typed value `time.Time` whose schema marks it optional and whose generated counterpart is `*time.Time`. An omitted timestamp decodes to the zero time and re-marshals as a fabricated `"0001-01-01T00:00:00Z"`, indistinguishable from a real value.

| Struct | Field | Wrapper | Wrapper type | `generated.*` type | `@required` in smithy? |
|---|---|---|---|---|---|
| `TimelineEvent` | `CreatedAt` | `timeline.go:20` | `time.Time` `` `json:"created_at"` `` | `*time.Time` | no |
| `WebhookDelivery` | `CreatedAt` | `webhooks.go:40` | `time.Time` `` `json:"created_at"` `` | `*time.Time` | no |
| `QuestionReminder` | `RemindAt` | `checkins.go:143` | `time.Time` `` `json:"remind_at"` `` | `*time.Time` | no |
| `ClientApprovalResponse` | `CreatedAt` | `client_approvals.go:68` | `time.Time` `` `json:"created_at"` `` | `*time.Time` | no |
| `ClientApprovalResponse` | `UpdatedAt` | `client_approvals.go:69` | `time.Time` `` `json:"updated_at"` `` | `*time.Time` | no |

Note the tags: **none carries `,omitempty`**. That is the exact shape of `SearchResult.CreatedAt`/`UpdatedAt`, which #615 did fix — a field tagged as if it were always present while the schema says otherwise. The key is emitted unconditionally, so absence surfaces as the fabricated instant with nothing in the wrapper source hinting the field was ever optional.

Controls, to show this is not "every timestamp": `generated.Webhook.CreatedAt`, `generated.Question.CreatedAt` and `generated.ClientApproval.CreatedAt` are all value `time.Time` (genuinely required), and `EverythingFile.CreatedAt`/`UpdatedAt` are already `*time.Time`. None of those appear above.

## How they were found

Neither grep nor the existing gates find these, because the wrapper source does not say the field is optional — the evidence lives in the generated client. Cross-reference the two by struct name and JSON tag:

- parse every non-test `go/pkg/basecamp/*.go` for struct fields typed exactly `time.Time`, keyed `(struct name, json tag)`
- parse `go/pkg/generated/client.gen.go` the same way
- report every wrapper field whose name-matched, tag-matched generated counterpart is `*time.Time`

At `a7354ae35` (#615's base) that scan yields **10 fields, 0 unpaired**: the 5 #615 fixes and the 5 above.

## Why no existing gate catches them

**`scripts/check-go-optional-pointers` cannot, at any scope.** It keys entirely on `,omitempty` (`OMITEMPTY` regex, `next unless m[3].match?(OMITEMPTY)`). All five above are tagged `json:"created_at"` with no options, so they are invisible to it even if its domain were widened to the wrapper. Widening it is independently a non-starter: pointed at the pre-#615 wrapper surface it reports **345** value-typed-`omitempty` fields, almost all deliberate `string`/`bool`/`int` flattening.

**`TestNoValueTypedOptionalTimestamps` (added by #615) cannot either**, for the same reason — it `continue`s on `!hasOmitempty(tag)`. It is a declaration guard reading only the wrapper's own tags, and the wrapper's tags are precisely what is wrong here.

**`scripts/check-wrapper-drift` is the right home but does not do it.** It already computes the `(wrapper, generated)` struct pairing from `*FromGenerated` converters, then compares JSON tag *names* only, never types. Teaching it to compare **nil-capability** — wrapper value type vs generated pointer — closes the whole class in the one place that already has both sides in hand.

Two known limits for whoever implements it:

1. `Notification` has no `*FromGenerated` converter (it is decoded by direct `json.Unmarshal`), so the converter-derived pairing does not cover it. A name-based fallback does.
2. The wrapper flattens optional `*string`/`*bool`/`*int` to value types **on purpose**. Any nil-capability comparison must be scoped to timestamps, or it re-derives the 345-violation wall. Timestamps are the principled carve-out: for a `string`, absent and `""` both marshal away harmlessly; for a `time.Time`, absent marshals as a wrong value.

## Why this is not a #615 amendment

Each of these is a silent breaking change with the same failure mode #615 documents — `t.IsZero()` keeps compiling against a pointer and panics on nil. #615 is converging for v0.13.0 with a written migration note covering its five. Adding five more fields across four more files, plus their converters and tests, is a second deliberate break that deserves its own migration note rather than a late amendment to one already under review.

## Related

- #562 — the timestamp instance that #615 closes. Its description table lists `TimelineEvent | CreatedAt, UpdatedAt | *time.Time`; `UpdatedAt` does not exist, and `CreatedAt` is one of the holdouts above rather than an example of the good convention.
- Six non-timestamp fields share the inert-`omitempty` shape (named structs, so `omitempty` never fires): `account_service.go:28-30`, `my_assignments.go:24-25`, `people.go:571`. Less harmful — a zero `{}` reads as obviously empty in a way `0001-01-01T00:00:00Z` does not — but the same root cause.

🤖 _Drafted with agent assistance; filed by @jeremy._

jeremy:
Closes #620. **Breaking** — five field types change from `time.Time` to `*time.Time`.

## The defect

| Struct | Field | File | Was | `generated.*` |
|---|---|---|---|---|
| `TimelineEvent` | `CreatedAt` | `timeline.go` | `time.Time` | `*time.Time` |
| `WebhookDelivery` | `CreatedAt` | `webhooks.go` | `time.Time` | `*time.Time` |
| `QuestionReminder` | `RemindAt` | `checkins.go` | `time.Time` | `*time.Time` |
| `ClientApprovalResponse` | `CreatedAt` | `client_approvals.go` | `time.Time` | `*time.Time` |
| `ClientApprovalResponse` | `UpdatedAt` | `client_approvals.go` | `time.Time` | `*time.Time` |

An omitted timestamp decoded to the zero time and re-marshaled as a fabricated `0001-01-01T00:00:00Z` — indistinguishable from a real value.

**None carried `,omitempty`, and it would not have helped.** `encoding/json`'s "empty" set has no entry for a struct, so `,omitempty` on a value-typed `time.Time` is inert: the key is emitted on every re-marshal.

Same class as #562/#615, which fixed five and left these five. Shipping the release with five fixed and five broken under an identical failure mode is the thing this avoids.

I re-derived the list from source rather than trusting the issue table — the cross-reference yields exactly these 5, 0 unpaired. Controls hold: `generated.Webhook.CreatedAt`, `generated.Question.CreatedAt` and `generated.ClientApproval.CreatedAt` are genuinely required value types and are untouched.

## Closing the class — the other half of the title

#620's title is "*and no gate can see them*". The evidence lives in the **generated** client, not the wrapper, so a guard must read both sides. Neither neighbouring gate can:

- **`TestNoValueTypedOptionalTimestamps`** (#615) keys on `,omitempty` and `continue`s without it. All five were tagged bare `json:"created_at"` — invisible. Its own doc comment named this gap and deferred it; this closes it.
- **`check-wrapper-drift`** has the `(wrapper, generated)` pairing but compares tag **names** only. Teaching it types would *still miss `WebhookDelivery`* — its header excludes the webhook-flavored shapes **by design** as a parallel representation, and `WebhookDelivery.CreatedAt` is one of the five. This is a limit the issue's suggested approach did not account for.

So `TestNoWrapperTimestampNarrowerThanGenerated` pairs by **struct name + json key**, which is blind to the converter tiers and covers all five. It fails loudly if the join matches nothing, so it cannot pass vacuously.

Scoped to timestamps deliberately: a nil-capability rule applied broadly reports ~345 intentionally flattened `*string`/`*bool`/`*int` fields. An absent string marshals away harmlessly; an absent `time.Time` marshals as a *wrong value*.

## Red proof, against un-fixed source

Un-fixed files swapped in with `git show origin/main:<path>`, my tests kept:

```
--- FAIL: TestOptionalTimestampsOmitWhenAbsent/TimelineEvent
    TimelineEvent.created_at was absent on the wire but re-marshaled as "0001-01-01T00:00:00Z"
--- FAIL: TestOptionalTimestampsOmitWhenAbsent/WebhookDelivery
--- FAIL: TestOptionalTimestampsOmitWhenAbsent/QuestionReminder
--- FAIL: TestOptionalTimestampsOmitWhenAbsent/ClientApprovalResponse  (created_at + updated_at)

--- FAIL: TestNoWrapperTimestampNarrowerThanGenerated
    ClientApprovalResponse.CreatedAt ... is time.Time but generated.* is *time.Time
    ClientApprovalResponse.UpdatedAt ...
    QuestionReminder.RemindAt ...
    TimelineEvent.CreatedAt ...
    WebhookDelivery.CreatedAt ...
REAL_EXIT=1
```

All five fail both the behavioural test and the new guard — including `WebhookDelivery`, the one no gate could reach. The positive control (`TestOptionalTimestampsRoundTripWhenPresent`) keeps a real timestamp round-tripping byte-identically.

## Verification

`make go-check go-check-wrapper-drift go-check-optional-pointers`, real exit written to a marker and grepped back:

```
ok  github.com/basecamp/basecamp-sdk/go/pkg/basecamp   9.083s
Wrapper drift check: 91 pairs walked, 1207 generated fields verified
MARKER_620_REAL_EXIT=0
```

## Migration

`t.CreatedAt.IsZero()` still compiles against a pointer and **panics on nil**. Callers nil-check, or use `basecamp.Deref`. This needs its own entry in `MIGRATING.md` (#642) — it re-opens that document's per-SDK tables and the A/B split, which I'll re-derive there against post-absorption `main`.

---
## Summary by cubic
Breaking: switch five Go wrapper timestamps from value time.Time to *time.Time so absent values stay nil and don’t re-marshal as 0001-01-01T00:00:00Z. Aligns wrappers with the generated client and closes #620.

- **Bug Fixes**
  - Changed to pointers: TimelineEvent.CreatedAt, WebhookDelivery.CreatedAt, QuestionReminder.RemindAt, ClientApprovalResponse.CreatedAt, ClientApprovalResponse.UpdatedAt. Converters now pass pointers through.
  - Added TestNoWrapperTimestampNarrowerThanGenerated to ensure wrapper timestamps aren’t narrower than the generated schema.
  - Updated tests to nil-check before calling IsZero/Equal on these pointer timestamps to avoid panics.

- **Migration**
  - CreatedAt/RemindAt/UpdatedAt are now pointers. Nil-check before use or use basecamp.Deref. Calling t.CreatedAt.IsZero() will panic if CreatedAt is nil.

<sup>Written for commit 71805573f64befbee0f0ed09066baf002e1a0599. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/basecamp/basecamp-sdk/pull/658?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/go/pkg/basecamp/checkins.go
+++ b/go/pkg/basecamp/checkins.go
@@ -144,10 +144,13 @@ type QuestionAnswer struct {
 
 // QuestionReminder represents a pending check-in reminder for the current user.
 type QuestionReminder struct {
-	GroupOn    string    `json:"group_on,omitempty"`
-	Question   Question  `json:"question"`
-	RemindAt   time.Time `json:"remind_at"`
-	ReminderID *int64    `json:"reminder_id,omitempty"`
+	GroupOn  string   `json:"group_on,omitempty"`
+	Question Question `json:"question"`
+	// RemindAt is optional — nil means the API did not send it. A value type
+	// here would fabricate 0001-01-01T00:00:00Z for an absent timestamp, which
+	// no consumer can tell from a real one.
+	RemindAt   *time.Time `json:"remind_at,omitempty"`
+	ReminderID *int64     `json:"reminder_id,omitempty"`
 }
 
 // QuestionNotificationSettings represents the current user's notification
@@ -1239,7 +1242,7 @@ func questionAnswerFromGenerated(ga generated.QuestionAnswer) QuestionAnswer {
 // questionReminderFromGenerated converts a generated QuestionReminder to our clean type.
 func questionReminderFromGenerated(gr generated.QuestionReminder) QuestionReminder {
 	r := QuestionReminder{
-		RemindAt:   deref(gr.RemindAt),
+		RemindAt:   gr.RemindAt,
 		ReminderID: gr.ReminderId,
 	}
 
--- a/go/pkg/basecamp/checkins_test.go
+++ b/go/pkg/basecamp/checkins_test.go
@@ -1361,7 +1361,9 @@ func TestCheckinsService_ListQuestionReminders(t *testing.T) {
 	if r1.GroupOn != "2022-10-28" {
 		t.Errorf("expected GroupOn '2022-10-28', got %q", r1.GroupOn)
 	}
-	if r1.RemindAt.IsZero() {
+	if r1.RemindAt == nil {
+		t.Error("expected RemindAt to be non-nil")
+	} else if r1.RemindAt.IsZero() {
 		t.Error("expected RemindAt to be non-zero")
 	}
 	if r1.ReminderID == nil || *r1.ReminderID != 123 {
--- a/go/pkg/basecamp/client_approvals.go
+++ b/go/pkg/basecamp/client_approvals.go
@@ -62,21 +62,24 @@ type ClientApproval struct {
 
 // ClientApprovalResponse represents a response to a client approval.
 type ClientApprovalResponse struct {
-	ID               int64     `json:"id"`
-	Status           string    `json:"status"`
-	VisibleToClients bool      `json:"visible_to_clients"`
-	CreatedAt        time.Time `json:"created_at"`
-	UpdatedAt        time.Time `json:"updated_at"`
-	Title            string    `json:"title"`
-	InheritsStatus   bool      `json:"inherits_status"`
-	Type             string    `json:"type"`
-	AppURL           string    `json:"app_url"`
-	BookmarkURL      string    `json:"bookmark_url"`
-	Parent           *Parent   `json:"parent,omitempty"`
-	Bucket           *Bucket   `json:"bucket,omitempty"`
-	Creator          *Person   `json:"creator,omitempty"`
-	Content          string    `json:"content"`
-	Approved         bool      `json:"approved"`
+	ID               int64  `json:"id"`
+	Status           string `json:"status"`
+	VisibleToClients bool   `json:"visible_to_clients"`
+	// CreatedAt and UpdatedAt are optional — nil means the API did not send
+	// them. Value types here would fabricate 0001-01-01T00:00:00Z for an absent
+	// timestamp, which no consumer can tell from a real one.
+	CreatedAt      *time.Time `json:"created_at,omitempty"`
+	UpdatedAt      *time.Time `json:"updated_at,omitempty"`
+	Title          string     `json:"title"`
+	InheritsStatus bool       `json:"inherits_status"`
+	Type           string     `json:"type"`
+	AppURL         string     `json:"app_url"`
+	BookmarkURL    string     `json:"bookmark_url"`
+	Parent         *Parent    `json:"parent,omitempty"`
+	Bucket         *Bucket    `json:"bucket,omitempty"`
+	Creator        *Person    `json:"creator,omitempty"`
+	Content        string     `json:"content"`
+	Approved       bool       `json:"approved"`
 }
 
 // ClientApprovalListResult contains the results from listing client approvals.
@@ -290,8 +293,8 @@ func clientApprovalFromGenerated(ga generated.ClientApproval) ClientApproval {
 			resp := ClientApprovalResponse{
 				Status:           deref(gr.Status),
 				VisibleToClients: deref(gr.VisibleToClients),
-				CreatedAt:        deref(gr.CreatedAt),
-				UpdatedAt:        deref(gr.UpdatedAt),
+				CreatedAt:        gr.CreatedAt,
+				UpdatedAt:        gr.UpdatedAt,
 				Title:            deref(gr.Title),
 				InheritsStatus:   deref(gr.InheritsStatus),
 				Type:             deref(gr.Type),
--- a/go/pkg/basecamp/optional_timestamps_test.go
+++ b/go/pkg/basecamp/optional_timestamps_test.go
@@ -78,6 +78,34 @@ func timestampCarriers() []timestampCarrier {
 			present: []byte(`{"id":1,"created_at":"2026-07-31T12:34:56Z","updated_at":"2026-07-30T01:02:03Z",` +
 				`"content":null,"description":null}`),
 		},
+		// #620: the five holdouts of the same class. Each was typed value
+		// time.Time with a bare `json:"..."` tag — no omitempty to even hint the
+		// field was optional — while its generated counterpart was *time.Time.
+		{
+			name:    "TimelineEvent",
+			decode:  unmarshalInto[TimelineEvent],
+			keys:    []string{"created_at"},
+			present: []byte(`{"id":1,"created_at":"2026-07-31T12:34:56Z","kind":"todo_created"}`),
+		},
+		{
+			name:    "WebhookDelivery",
+			decode:  unmarshalInto[WebhookDelivery],
+			keys:    []string{"created_at"},
+			present: []byte(`{"id":1,"created_at":"2026-07-31T12:34:56Z"}`),
+		},
+		{
+			name:    "QuestionReminder",
+			decode:  unmarshalInto[QuestionReminder],
+			keys:    []string{"remind_at"},
+			present: []byte(`{"remind_at":"2026-07-31T12:34:56Z"}`),
+		},
+		{
+			name:   "ClientApprovalResponse",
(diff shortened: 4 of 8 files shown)


jeremy:
**B1 — confirmed, fixed.** Three assertions in this package's own tests dereferenced the newly-pointerized timestamps with no nil guard:

- `go/pkg/basecamp/checkins_test.go:1364` — `r1.RemindAt.IsZero()`
- `go/pkg/basecamp/webhooks_test.go:177` — `delivery.CreatedAt.IsZero()`
- `go/pkg/basecamp/timeline_test.go:81` — `!event.CreatedAt.Equal(expectedTime)`

All three compile because Go auto-dereferences for a value receiver, and all three pass only because the fixtures happen to carry the key. Drop it and they panic, taking the test binary down instead of failing with a message naming the field — the exact hazard this PR documents for consumers. Each now requires non-nil with its own message and then asserts what it asserted before; `timeline_test.go` also prints `*event.CreatedAt` rather than the pointer.

I swept the rest of the package for the same shape rather than fixing only the three named. `rg` over `go/pkg/basecamp/*_test.go` for `.CreatedAt.` / `.UpdatedAt.` / `RemindAt.` turns up plenty more, but every other hit is on a struct whose field is still value-typed `time.Time` — `ClientApproval` (not `ClientApprovalResponse`), `Campfire`, `Card`, `Boost`, and so on. The only test touching `ClientApprovalResponse` directly (`client_approvals_test.go:177`) never reads its timestamps. Three is the complete set.

**Verification** (exit code written into the log under a marker and grepped back):

```
make go-check go-check-wrapper-drift go-check-optional-pointers
MARKER_GO620_EXIT=0
ok  github.com/basecamp/basecamp-sdk/go/pkg/basecamp  9.092s
Wrapper drift check: 91 pairs walked, 1207 generated fields verified
```

jeremy:
Merging on independent adversarial review; both bot reviewers were unavailable this session.

The review re-derived the field list from scratch rather than trusting mine — a standalone program replicating the (struct name, json key) join against both trees. Against `origin/main`: exactly five violations, the five this PR fixes. Against this branch: zero. There is no sixth.

It also measured the guard's reach, which is better than I claimed: **119 wrapper timestamp fields, 118 paired, and the single unpaired one (`Todo.CompletedAt`) is already a pointer** — so there are currently zero value-typed wrapper timestamps invisible to the join. It fails loudly on an empty join and is genuinely exercised at 118 pairs.

Two things fixed since review:
- Added the `breaking` label. It was missing, and `.github/release.yml` derives the breaking-changes section solely from that label. The `API Compatibility` check cannot substitute — test.yml:859 is `# Don't fail the build, just warn`, and it only diffs `./pkg/basecamp`.
- Three unguarded dereferences left in this package's own tests (`checkins_test.go:1364`, `webhooks_test.go:177`, `timeline_test.go:81`). All compiled unchanged because Go auto-dereferences for a value receiver, and passed only because the fixtures happen to supply the key — the precise hazard this PR documents for consumers. Each now requires non-nil with its own message before asserting. A sweep confirms three was the complete set; every other `.CreatedAt.`/`.UpdatedAt.` hit is on a still-value-typed struct.

One finding deliberately not fixed here, filed as #662: `AuthorizationInfo.ExpiresAt` is a value `FlexTime` whose decoder maps null to the zero time, so an absent expiry re-marshals as a fabricated `0001-01-01T00:00:00Z` — the same failure class, on a credential-expiry field. Both timestamp guards are blind to it because `isValueTime` matches only the literal `time.Time` selector, so any named type embedding `time.Time` is invisible. It also has no generated counterpart to pair against, being decoded straight from Launchpad. That needs a guard change and a question answered about whether Launchpad ever omits the field, neither of which belongs in a freeze.

Issue #576, opened 2026-08-01
Python, Ruby and TypeScript merge-safe composites forward a non-string GET field into the full-replace PUT


jeremy:
Found while fixing the same defect class in the Kotlin Todolists composite (#574, Codex's third pass). **This affects shipped code on `main`** — the Todos composites in Python, Ruby and TypeScript, and the Cards composites in Python, Ruby and TypeScript.

> **Body corrected 2026-08-01.** The first version of this issue called TypeScript "structurally safe" and scoped remediation to Python and Ruby, and it missed Ruby Cards entirely. Both were wrong; see the corrections below and the comment thread for the audit trail. The original reasoning is preserved in the "How the wrong verdict happened" section, because the mistake is instructive.

## The shape

A merge-safe `update`/`edit` GETs the record, reads each writable field, and PUTs the **full** representation back. If the read step coerces or forwards a non-string value instead of rejecting it, that value is written over the real one — on a call that never mentioned the field. The endpoint is full-replace, so this is data loss or corruption, not a no-op.

There are two distinct failure modes, and testing for only the first is what produced this issue's original wrong verdict:

- **Erasure** — a falsey non-string is coalesced to `""`, wiping the field.
- **Corruption** — a truthy or non-coalesced non-string is forwarded verbatim, writing a number, boolean, array or object where a string belongs.

## Affected

**Python** (`python/src/basecamp/services/todos.py:22-34`) — `todo.get("description") or ""`:

| GET value | sent as | |
|---|---|---|
| `[]`, `{}`, `0`, `False` | `''` | **erased** |
| `42`, `True` | verbatim | **corrupted** |

**Ruby Todos** (`ruby/lib/basecamp/services/todos_extensions.rb:88-98`) — `todo["description"] || ""`:

| GET value | sent as | |
|---|---|---|
| `false` | `""` | **erased** |
| `[]`, `{}`, `0`, `42`, `true` | verbatim | **corrupted** |

`||` treats only `nil`/`false` as falsy, so Ruby erases in one case and corrupts in the rest.

[...]

PR #660, opened 2026-08-04
Kotlin: stop coercing a wrong-typed scalar into a String (#598)

### LINKED ISSUE #598: Kotlin coerces a wrong-typed scalar into a String, so the merge-safe composites write it back
Found by the shared kill fixtures added in #597, which is exactly what those fixtures exist for: five consecutive language-specific review passes on this defect class missed it.

## The claim that is wrong

#576 lists Kotlin Todos and Cards under **"Not affected, and why"**:

> **Kotlin Todos and Cards** — `todos.get()` returns `Todo`, `cards.get()` returns `Card`; kotlinx.serialization rejects a wrong-typed field at decode.

That is only true for a **structural** mismatch (an array or object where a string belongs). It is **not** true for a scalar. `BasecampClient.kt:156` configures the client-wide decoder as:

```kotlin
internal val json: Json = Json {
    ignoreUnknownKeys = true
    isLenient = true
    coerceInputValues = true
}
```

`isLenient = true` makes kotlinx.serialization accept a JSON number or boolean where a `String` is declared, coercing it. So `"description": 42` decodes into `Todo.description` as `"42"`, the merge-safe composite reads a perfectly ordinary `String`, and PUTs it back to a full-replace endpoint — on a call that never mentioned the field.

## Proof, on the wire

Run against the conformance kill fixtures with their assertions temporarily flipped from `errorRaised` + `requestCount: 1` to `requestCount: 2` plus a `requestBody` pin on the second request. All four **PASSED**, meaning the second request really happened and really carried the coerced value:

```
PASS: update-kill … (requestCount 2, requestBody description == "42")
PASS: update-kill … (requestCount 2, requestBody description == "false")
PASS: update-kill … (requestCount 2, requestBody due_on == "false")
PASS: update-kill … (requestCount 2, requestBody due_on == "42")
```

So, driving `account.todos.update(todoId, content = "New title")` and `account.cards.update(cardId, title = "Renamed card")`:

| GET value | Kotlin PUTs | |
|---|---|---|
| `"description": 42` | `"description": "42"` | corrupted |
| `"description": false` | `"description": "false"` | corrupted |
| `"due_on": 42` | `"due_on": "42"` | corrupted |
| `"due_on": false` | `"due_on": "false"` | corrupted |
| `"description": []` / `{}` | — | correctly refused (structural mismatch) |
| `"due_on": []` / `{}` | — | correctly refused |

Severity matches #576: **latent, not actively firing.** It needs BC3 to return a non-string, which it does not do today. It is filed because the guarantee these methods advertise is "your unmentioned fields are safe", and in Kotlin that currently rests on the server never misbehaving.

## Why #597's pattern cannot fix it

#597 fixes Python, Ruby and TypeScript with a per-composite guard between the GET and the field read. That works there because the composite sees the raw `dict`/`Hash`/erased-type value. In Kotlin the coercion has **already happened** by the time the composite runs — it only ever sees a `String`, with no way to know whether the wire carried `42` or `"42"`. A guard in `TodosService`/`CardsService` is structurally incapable of catching this.

The real options, all bigger than a containment fix:

1. **Drop `isLenient` (and reconsider `coerceInputValues`) client-wide.** Correct in principle and the smallest diff, but the blast radius is every response the SDK decodes, so it needs its own drift/conformance sweep — a lenient decoder may be silently absorbing other BC3 shapes today.
2. **Per-field strict serializers** on the writable string properties of the composite-backed models. Narrow, but hand-written on generated models, so it needs generator support to not rot.
3. **Fold it into #578**, the generated validating layer. If that layer owns "reject a wrong-typed field", Kotlin's decoder config stops being the thing standing between a malformed body and a full-replace PUT.

Option 1 is the one worth pricing first, because `isLenient` is doing something for every model in the SDK and nobody currently knows what.

## Interim coverage

The kill fixtures shipped in #597 use array/object values, which kotlinx.serialization *does* reject, so Kotlin passes them honestly today — they pin the structural half of the contract in all six SDKs. The scalar half is **unguarded in Kotlin and only in Kotlin**, and there is no fixture holding that line, because no single value discriminates for both Kotlin (which needs a structural shape) and TypeScript Cards (which needs a falsey scalar). Adding a Kotlin-only kill case belongs with the fix.

Related: #576 (the three languages fixed in #597), #578 (generated validating layer), #574 (where this class was first hunted).

jeremy:
Closes #598.

Kotlin's client-wide `Json` carried `isLenient = true`. That relaxes RFC-4627 far enough to read a JSON number or boolean into a declared `String`, so `"description": 42` decoded to `"42"` — and the merge-safe composites then PUT that fabricated value back to a full-replace endpoint on a call that never mentioned the field. The coercion happens inside the decoder, so the per-composite guard that fixed Python, Ruby and TypeScript in #597 is structurally incapable of catching it: by the time the composite runs, the value is an ordinary `String` with no way to know whether the wire carried `42` or `"42"`.

The fix is one line: drop `isLenient`, keep `coerceInputValues`.

## The flag attribution — `isLenient`, not `coerceInputValues`

The issue text and the release plan disagreed about the cause. They are not interchangeable, and removing the wrong one ships a no-op as a fix. The issue was right.

`kotlin/sdk/src/commonTest/kotlin/com/basecamp/sdk/DecoderStrictnessTest.kt` is new and pins the semantics of each flag in isolation, so the answer lives in the tree rather than in anyone's recollection:

- `lenientCoercesNumberIntoString` / `lenientCoercesBooleanIntoString` — `isLenient` alone produces `"42"` and `"false"` from `42` and `false`. **This is the defect.**
- `coerceInputValuesDoesNotCoerceAScalarType` — `coerceInputValues` alone **throws** on `{"description": 42}`. It is not implicated, and dropping it would have been a no-op fix.
- `coerceInputValuesRewritesNullToTheDeclaredDefault` / `withoutCoerceInputValuesAnExplicitNullIsRejected` — what `coerceInputValues` is actually for: an explicit `null` into a non-nullable property with a default. That is why it stays.
- `nullableFieldsAcceptExplicitNullEverywhere` — the shape the SDK's own generated models use, so keeping or dropping `coerceInputValues` is invisible to them.
- `structuralMismatchIsRejectedEvenWhenLenient` — arrays and objects were always refused. That is the half #576 correctly called safe, and it stays that way.

## The tripwire flipped

`TodolistsServiceTest` carried a test that asserted the defect on purpose, and said so in its own doc comment:

> This is a CROSS-SERVICE gap (#576) — the Documents composite has it too — not something a todolist-specific guard should paper over [...] This test records the exact behavior **so closing #576 flips it visibly.**

It asserted `assertEquals(listOf("GET", "PUT"), capture.methods)` and `assertEquals("42", capture.putBody!!["description"])` — the coerced value landing on the wire.

It is now `updateRefusesABareScalarDescriptionBeforeWriting` and asserts the refusal: no PUT is issued, and the caller gets a statusless, non-retryable `BasecampException.Api` carrying a hint — for both a number (`42`) and a boolean (`false`), because `isLenient` accepted both and each produced a differently-shaped fabrication. That flip is the evidence the fix landed, on exactly the terms the original author asked for.

**Todolists is a third affected service**, beyond the Todos and Cards that #576 and #598 named. Documents and Schedules are the same composite shape and their comments are corrected here too.

## Conformance: a bare-scalar kill case, and it needs no per-language skip

#598 anticipated a Kotlin-only fixture, since "no single value discriminates for both Kotlin (which needs a structural shape) and TypeScript Cards (which needs a falsey scalar)." That was true of the code as it stood. **The decoder fix removes the asymmetry**, so the fixture is better than anticipated: with `isLenient` gone, all six SDKs refuse a non-string scalar — Go, Kotlin and Swift in their model decoders (`json.Unmarshal`, kotlinx.serialization, `Codable`), and TypeScript, Python and Ruby in the hand-written `writableString`/`write_string` guards, whose non-string branch always covered `42`. So `conformance/tests/todos_write.json` gains

> `update-kill: a bare-scalar description is refused before the full-replace PUT`

as an ordinary six-runner case with **zero entries added to any runner's skip map**. Confirmed executing, not silently skipped, in all six: Go/Kotlin/Ruby/Python/Swift each print a PASS line, and TypeScript (which is non-verbose in the suite) was checked directly with `npx vitest run -t 'bare-scalar' --reporter=verbose` → `✓ conformance/todos_write.json > update-kill: a bare-scalar description is refused before the full-replace PUT`.

`conformance/check_kill_case_controls.py` passes: the new body is the existing `update-merge` control's body with exactly one field perturbed, same as its array and object siblings.

### It is a kill case, not decoration

Restoring `isLenient = true` locally and re-running `make conformance-kotlin`:

```
  FAIL: update-kill: a bare-scalar description is refused before the full-replace PUT
        Expected the call to fail, but it succeeded
Passed: 175, Failed: 1, Skipped: 1, Total: 177
MARKER_KT598_RED_EXIT=2
```

Reverted immediately; the committed tree has no `isLenient` assignment (`git grep isLenient` under `kotlin/sdk/src/commonMain` returns comment lines only).

## Prose the fix falsified

Corrected only where the claim is now false — comments describing the scalar hole as an open cross-service gap, several of which also named the wrong flag pair:

- `TodolistsServiceTest.updateRefusesAMalformedDescriptionBeforeWriting` — said arrays/objects were used *because* bare scalars get coerced; now cross-references the sibling.
- `DocumentsServiceTest`, `SchedulesServiceTest` — same "An ARRAY, not a scalar" rationale.
- `DocumentsService.kt` (×2), `SchedulesService.kt`, `TodolistsService.kt` — "(The client-wide `coerceInputValues`/`isLenient` scalar hole means a bare JSON scalar is coerced rather than rejected [...] not something this composite can close.)"
- `conformance/tests/todos_write.json` — the two existing kill-case descriptions asserted a scalar "would NOT discriminate in Kotlin".

`SPEC.md:1091` mentions `coerceInputValues = true` in the `FlexibleIntSerializer` discussion and is still accurate — that flag is unchanged — so it is left alone. Same for `ReplayRunner.kt`, whose stricter replay `Json` is unaffected.

## Verification

Full suite run twice on the final tree, exit code captured via a marker written into the log and grepped back rather than trusted from a mid-run "BUILD SUCCESSFUL":

```
make kt-check conformance
MARKER_KT598_REAL_EXIT=0    (run 1)
MARKER_KT598_REAL_EXIT=0    (run 2)
```

Kotlin SDK checks passed and every runner was green. An earlier revision of this description claimed `Passed: 176, Failed: 0, Skipped: 1, Total: 177` "in each runner" — that is Kotlin's and Swift's tally, not a shared one. The runners do not report a common shape, and their skip counts differ. The real per-runner numbers, from a fresh `make conformance` on the final tree:

| Runner | Passed | Failed | Skipped | Total |
|---|---|---|---|---|
| Go | 175 | 0 | 2 | 177 |
| Kotlin | 176 | 0 | 1 | 177 |
| TypeScript | 219 | 0 | 2 | 221 |
| Ruby | 166 | 0 | 11 | 177 |
| Python | 177 | 0 | 0 | 177 |
| Swift | 176 | 0 | 1 | 177 |

```
MARKER_CONFORMANCE_EXIT=0
```

TypeScript counts differently — it runs two suites, so 221 is not comparable to the others' 177. No test other than the deliberate tripwire needed flipping.

---
## Summary by cubic
Stops Kotlin from coercing wrong-typed scalars into Strings by removing `isLenient` from the client-wide `Json`. Numbers/booleans are now rejected before any write, aligning Kotlin with the other SDKs and closing #598.

- **Bug Fixes**
  - Removed `isLenient` from the client-wide `Json`; kept `coerceInputValues` for explicit nulls.
  - Strengthened tests: `DecoderStrictnessTest` pins each flag and now decodes through `BasecampClient.json` to guard the actual config.
  - Todolists: flipped the scalar test to assert refusal before PUT; updated Documents and Schedules tests and comments.
  - Cards: added a test to refuse a bare-scalar `due_on` from the server response.
  - Conformance: added a bare-scalar kill case to `conformance/tests/todos_write.json`; passes in all six SDKs with no skips.

<sup>Written for commit 0c3b16a9f74b06ec5d7c0e51279688ebc58ea4cc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/basecamp/basecamp-sdk/pull/660?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/conformance/tests/todos_write.json
+++ b/conformance/tests/todos_write.json
@@ -550,7 +550,7 @@
   },
   {
     "name": "update-kill: an array description is refused before the full-replace PUT",
-    "description": "The merge-safe update GETs the todo, reads every writable field, and PUTs the FULL representation back — so a value read is a value written, on a call that only set the content. A malformed one must be refused BEFORE the PUT, which is exactly what requestCount: 1 pins: the moment a second request appears, the malformed value has already landed on a full-replace endpoint.\n\nThis is the CORRUPTION half of #576. Neither `or \"\"` (Python) nor `|| \"\"` (Ruby) nor `?? \"\"` (TypeScript) coalesces a non-empty array, so on main all three PUT `\"description\": [\"x\"]` to the todo — an array where the rich-text body belongs.\n\nThe kill case for #576, and the reason it lives here rather than in six per-language test files: the defect survived five consecutive review passes because each pass fixed one instance. A shared fixture catches every instance at once, in every runner, permanently — and it did: writing it surfaced the Kotlin scalar-coercion hole above, which five language-specific passes had missed.\n\nAll six SDKs must refuse it, by two different mechanisms, which is why the value under test is a JSON array/object rather than a scalar. Go, Kotlin and Swift decode into a typed model: json.Unmarshal, kotlinx.serialization and Codable all reject a structural mismatch. TypeScript, Python and Ruby have no decoder at all — schema.d.ts is erased at build time, and the generated Python and Ruby services return an untyped dict/Hash — so they carry an explicit guard. A scalar (42, false) would NOT discriminate in Kotlin: its client-wide Json { isLenient = true } coerces one into a String field, so the composite decodes \"42\" and writes it back. That is the same defect one layer down and is tracked separately; it cannot be fixed by a per-composite guard, because by the time the composite runs the coercion has already happened.\n\nerrorRaised rather than errorType: a hand-written guard raises api_error while a decoder raises its own language's decode failure, so no single canonical code is assertable. What all six agree on is that the call fails at all — which, paired with requestCount: 1, is the whole contract.",
+    "description": "The merge-safe update GETs the todo, reads every writable field, and PUTs the FULL representation back — so a value read is a value written, on a call that only set the content. A malformed one must be refused BEFORE the PUT, which is exactly what requestCount: 1 pins: the moment a second request appears, the malformed value has already landed on a full-replace endpoint.\n\nThis is the CORRUPTION half of #576. Neither `or \"\"` (Python) nor `|| \"\"` (Ruby) nor `?? \"\"` (TypeScript) coalesces a non-empty array, so on main all three PUT `\"description\": [\"x\"]` to the todo — an array where the rich-text body belongs.\n\nThe kill case for #576, and the reason it lives here rather than in six per-language test files: the defect survived five consecutive review passes because each pass fixed one instance. A shared fixture catches every instance at once, in every runner, permanently — and it did: writing it surfaced the Kotlin scalar-coercion hole above, which five language-specific passes had missed.\n\nAll six SDKs must refuse it, by two different mechanisms. Go, Kotlin and Swift decode into a typed model: json.Unmarshal, kotlinx.serialization and Codable all reject a structural mismatch. TypeScript, Python and Ruby have no decoder at all — schema.d.ts is erased at build time, and the generated Python and Ruby services return an untyped dict/Hash — so they carry an explicit guard. The value under test is an array/object rather than a scalar for a reason that has since been removed: a scalar (42, false) did NOT discriminate in Kotlin, whose client-wide Json { isLenient = true } coerced one into a String field, so the composite decoded \"42\" and wrote it back. That was the same defect one layer down, and no per-composite guard could close it — by the time the composite ran, the coercion had already happened. #598 dropped isLenient, and the bare-scalar sibling below now pins the scalar half in all six runners.\n\nerrorRaised rather than errorType: a hand-written guard raises api_error while a decoder raises its own language's decode failure, so no single canonical code is assertable. What all six agree on is that the call fails at all — which, paired with requestCount: 1, is the whole contract.",
     "operation": "UpdateTodo",
     "method": "PUT",
     "path": "/todos/{todoId}",
@@ -714,7 +714,7 @@
   },
   {
     "name": "update-kill: an empty-object description is refused, not coalesced to empty",
-    "description": "The ERASURE half of #576, and the half whose absence produced this issue's original wrong verdict. Python's `or \"\"` turns every falsey non-string — {} included — into \"\", and on a full-replace endpoint that empty value is written back over the real description: the field the composite exists to preserve, wiped by a call that only set the content. Ruby and TypeScript reach the same endpoint by the other road and forward `{}` verbatim. Testing for corruption alone would miss the erasure; testing for erasure alone is what let TypeScript be called structurally safe.\n\nThe kill case for #576, and the reason it lives here rather than in six per-language test files: the defect survived five consecutive review passes because each pass fixed one instance. A shared fixture catches every instance at once, in every runner, permanently — and it did: writing it surfaced the Kotlin scalar-coercion hole above, which five language-specific passes had missed.\n\nAll six SDKs must refuse it, by two different mechanisms, which is why the value under test is a JSON array/object rather than a scalar. Go, Kotlin and Swift decode into a typed model: json.Unmarshal, kotlinx.serialization and Codable all reject a structural mismatch. TypeScript, Python and Ruby have no decoder at all — schema.d.ts is erased at build time, and the generated Python and Ruby services return an untyped dict/Hash — so they carry an explicit guard. A scalar (42, false) would NOT discriminate in Kotlin: its client-wide Json { isLenient = true } coerces one into a String field, so the composite decodes \"42\" and writes it back. That is the same defect one layer down and is tracked separately; it cannot be fixed by a per-composite guard, because by the time the composite runs the coercion has already happened.\n\nerrorRaised rather than errorType: a hand-written guard raises api_error while a decoder raises its own language's decode failure, so no single canonical code is assertable. What all six agree on is that the call fails at all — which, paired with requestCount: 1, is the whole contract.",
+    "description": "The ERASURE half of #576, and the half whose absence produced this issue's original wrong verdict. Python's `or \"\"` turns every falsey non-string — {} included — into \"\", and on a full-replace endpoint that empty value is written back over the real description: the field the composite exists to preserve, wiped by a call that only set the content. Ruby and TypeScript reach the same endpoint by the other road and forward `{}` verbatim. Testing for corruption alone would miss the erasure; testing for erasure alone is what let TypeScript be called structurally safe.\n\nThe kill case for #576, and the reason it lives here rather than in six per-language test files: the defect survived five consecutive review passes because each pass fixed one instance. A shared fixture catches every instance at once, in every runner, permanently — and it did: writing it surfaced the Kotlin scalar-coercion hole above, which five language-specific passes had missed.\n\nAll six SDKs must refuse it, by two different mechanisms. Go, Kotlin and Swift decode into a typed model: json.Unmarshal, kotlinx.serialization and Codable all reject a structural mismatch. TypeScript, Python and Ruby have no decoder at all — schema.d.ts is erased at build time, and the generated Python and Ruby services return an untyped dict/Hash — so they carry an explicit guard. The value under test is an array/object rather than a scalar for a reason that has since been removed: a scalar (42, false) did NOT discriminate in Kotlin, whose client-wide Json { isLenient = true } coerced one into a String field, so the composite decoded \"42\" and wrote it back. That was the same defect one layer down, and no per-composite guard could close it — by the time the composite ran, the coercion had already happened. #598 dropped isLenient, and the bare-scalar sibling below now pins the scalar half in all six runners.\n\nerrorRaised rather than errorType: a hand-written guard raises api_error while a decoder raises its own language's decode failure, so no single canonical code is assertable. What all six agree on is that the call fails at all — which, paired with requestCount: 1, is the whole contract.",
     "operation": "UpdateTodo",
     "method": "PUT",
     "path": "/todos/{todoId}",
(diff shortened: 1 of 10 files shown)


jeremy:
Absorbed the review. C1 and C3 were right; C2 was right that Cards had no coverage but wrong about the shape it should take, and I have evidence rather than an opinion on that.

**C1 — the comment was writing a cheque the test did not cover.** `BasecampClient.kt` claimed "DecoderStrictnessTest pins both behaviours so this comment cannot drift from the semantics". It cannot: `DecoderStrictnessTest.kt:34-36` builds its own three `Json` instances and the file never mentions `BasecampClient`. It pins kotlinx.serialization's flag semantics, not this SDK's configuration.

Reproduced the reviewer's proof: restoring `origin/main`'s `BasecampClient.kt` (`isLenient = true`) and running the file alone was **green, 10/10**.

`json` is `internal` and `commonTest` is the associated compilation, so it is reachable. Added `theClientsOwnDecoderRefusesAWrongTypedScalar`, which builds a real client via `testBasecampClient` and decodes `{"description": 42}` / `{"description": false}` through `client.json` — plus an explicit null into a non-nullable probe, so `coerceInputValues` staying on is pinned too.

RED, with `origin/main`'s `BasecampClient.kt` swapped in:

```
DecoderStrictnessTest[jvm] > theClientsOwnDecoderRefusesAWrongTypedScalar()[jvm] FAILED
    org.opentest4j.AssertionFailedError at DecoderStrictnessTest.kt:126
10 tests completed, 1 failed
MARKER_C1_RED_EXIT=1
```

It is the *only* failure — which is the reviewer's point restated as a number: the nine pre-existing cases stay green against the un-fixed client.

GREEN, on this branch:

```
MARKER_C1_GREEN_EXIT=0
BUILD SUCCESSFUL
```

(The swap was done with `git show origin/main:<path>` and restored the same way, never `git stash` — this worktree shares a ref store with ~30 others.)

Narrowed the six `assertFailsWith<Exception>` to `SerializationException`. All six pass narrowed, so every one of those cases really does throw it; none needed to stay broad. Reworded the comment to name the new test rather than the file.

**C2 — Cards genuinely had no coverage, but the requested shape is no longer constructible.** The ask was to mirror `updateRefusesABareScalarDescriptionBeforeWriting`: GET returns a bare-scalar `due_on`, `cards.update` refuses before the PUT, assert no PUT was issued. `cards.update` issues no GET. #647 — `46b7f8225`, the parent commit of this branch — deleted the preservation GET:

> With omission now meaning "unchanged", the read-modify-write preservation GET has nothing left to protect, so it goes too. update is a single PUT in every case.

and removed the three Cards kill cases from `conformance/tests/cards_write.json` in the same commit, for exactly this reason:

```
-    "name": "update-kill: an array due_on is refused before the replacement PUT",
-    "name": "update-kill: an empty-object due_on is refused, not coerced or dropped",
-    "name": "update-kill: a date-shaped array due_on is refused where the format check is blind",
```

`CardsService.kt` is now a straight delegation to `updateVerbatim`. There is no read-back to refuse and no PUT to suppress; the first request the method makes *is* the PUT. #598's table describes the pre-#647 code.

What does survive is the response decode, and it is the same shared decoder: `updateVerbatim` ends in `json.decodeFromString<Card>(body)`, and `Card.dueOn` is `String?`. Under `isLenient` a bare-scalar `due_on` coming back off the wire was coerced and handed to the caller as an ordinary String. So I covered that instead — `updateRefusesABareScalarDueOnFromTheWire`, both `42` and `false`, asserting the refusal and that exactly one PUT was issued.

It goes RED against the un-fixed client, and the failure message is the #598 defect verbatim:

```
CardsServiceTest[jvm] > updateRefusesABareScalarDueOnFromTheWire()[jvm] FAILED
7 tests completed, 1 failed
MARKER_C2_RED_EXIT=1

a bare-scalar due_on (42) must be refused, not rendered as text.
Expected an exception of class kotlinx.serialization.SerializationException to be thrown,
but was completed successfully with the result: <Card(id=42, ..., dueOn=42, ...)>
```

`dueOn=42` — a fabricated String on a `String?` field, exactly what the issue reported. GREEN on this branch: `MARKER_C2_GREEN_EXIT=0`.

If you would rather this were pinned in conformance than in a unit test, say so and I will add it there; I kept it local because the equivalent conformance cases were deliberately retired by #647.

**C3 — the numbers were wrong, corrected in the body.** "Passed: 176, Failed: 0, Skipped: 1, Total: 177 in each runner" is Kotlin's and Swift's tally, presented as if shared. Re-ran `make conformance` on the final tree:

| Runner | Passed | Failed | Skipped | Total |
|---|---|---|---|---|
| Go | 175 | 0 | 2 | 177 |
| Kotlin | 176 | 0 | 1 | 177 |
| TypeScript | 219 | 0 | 2 | 221 |
| Ruby | 166 | 0 | 11 | 177 |
| Python | 177 | 0 | 0 | 177 |
| Swift | 176 | 0 | 1 | 177 |

Matches the reviewer's figures on every runner. TypeScript's 221 is two suites, not comparable to the others' 177.

**Verification** (exit codes written into the logs under markers and grepped back):

```
make kt-check      MARKER_KTCHECK_EXIT=0
make conformance   MARKER_CONFORMANCE_EXIT=0
```

Incidental confirmation from the Ruby conformance output, which bears on #656's README fix: `PASS: DownloadURL retries on 503 at the auth'd first hop` / `PASS: DownloadURL does not retry hop 1 on 500`.

jeremy:
Merging on independent adversarial review; both bot reviewers were unavailable this session.

The headline finding is one the issue and the release plan disagreed about, settled by experiment rather than argument: **`isLenient` is the culprit, not `coerceInputValues`.** A permutation probe pins it — `isLenient` reads `42` into a declared String; `coerceInputValues` only rewrites an explicit null to a declared default and has no opinion on scalar type. Dropping the flag the plan named would have shipped a no-op as a fix.

Blast radius was swept rather than assumed, since this changes a client-wide decoder during a freeze: 229 `.kt` sources, 71 conformance fixtures and 124 `spec/fixtures/**` contain **zero** unquoted keys, single quotes, `NaN`/`Infinity`, or quoted-numbers-into-numerics reaching the client decoder. The only lenient-shaped JSON in the tree is the kill fixture this PR adds. Apparent counterexamples resolve: `FieldKeyed422Test`'s `"error_description": 42` goes through a tree parse whose `stringMember` rejects non-strings, and the `Flexible*SerializerTest` literals call `decodeJsonElement()` themselves. `git log -S` puts `isLenient` in the original SDK commit, not in response to any BC3 shape — it was never a deliberate accommodation.

Two things fixed after review:
- `DecoderStrictnessTest` built its own `Json` instances and never touched `BasecampClient`, so it could not see this fix — proven by running it green against the un-fixed client — while the client comment claimed it "cannot drift from the semantics". Added `theClientsOwnDecoderRefusesAWrongTypedScalar`, which decodes through the instance the client actually hands every service. Verified red (that test the only failure) then green. Also narrowed six `assertFailsWith<Exception>` to `SerializationException`.
- The conformance numbers in the body were wrong and are now the measured per-runner figures.

One correction to my own request: I asked for a Cards test asserting refusal before the PUT. That is not constructible — #647, this branch's parent, deleted the Cards preservation GET and removed those `update-kill` cases as unreachable, so `cards.update` is now a straight delegation to `updateVerbatim` with no GET. Covered the path that does survive (the response decode, `Card.dueOn` being `String?` through the shared decoder) with its own red/green instead.

Worth recording: the decoder fix removed the asymmetry #598 assumed. The issue expected a Kotlin-only kill case needing five skips, because no single value discriminated for both Kotlin and TypeScript Cards. With Kotlin fixed a bare scalar discriminates in all six, so the case landed as an ordinary six-runner fixture with **zero** skip entries — the scalar half of the contract is now pinned everywhere, which the issue described as unguarded in Kotlin and only in Kotlin.

Not labelled `breaking`, following #597 — the identical fix for TypeScript, Python and Ruby — which shipped without it.

Issue #633, opened 2026-08-04
types.FlexibleTime is asymmetric: accepts a bare date, marshals it back as a midnight timestamp


jeremy:
Found while implementing the ScheduleEntries triad (#632), reported rather than absorbed — it predates that work and deserves its own fix.

## The asymmetry

`types.FlexibleTime` **accepts** a bare date (`"2026-08-04"`) on unmarshal, but **marshals it back** as a midnight timestamp (`"2026-08-04T00:00:00Z"`).

BC3 then re-parses that timestamp in the **account's timezone**. For any account west of UTC, midnight-UTC is the *previous day* locally — so an all-day entry read and written back **lands a day early**.

That is a silent data corruption on the read-modify-write path, which is exactly what a merge-safe composite does.

## Why the triad did not absorb it

Go's composites currently read **raw response bytes** to sidestep the round-trip entirely. That is a correct local workaround, but it is a workaround: any future code path that unmarshals into `FlexibleTime` and marshals back inherits the bug.

## What a fix needs

`FlexibleTime` must remember which form it parsed and re-emit the same one — a bare date round-trips as a bare date, a timestamp as a timestamp. That is the only shape that makes read-modify-write safe.

**Red-proof it in both directions**, since a one-directional test would pass today:

1. Unmarshal `"2026-08-04"` → marshal → assert the output is `"2026-08-04"`, not `"2026-08-04T00:00:00Z"`. This must **fail** before the fix.
2. Unmarshal a real timestamp → marshal → assert it is unchanged. This guards against "fix" by always emitting a bare date.

Worth also checking the sibling flexible types (`types.FlexInt`, `types.FlexibleInt64`, `types.Date`) for the same shape — a type that widens on input and narrows on output is a general hazard, and this is unlikely to be the only instance.

jeremy:
Fix in #852: `types.FlexibleTime` remembers whether it parsed a bare date and marshals that form back, red-proofed in both directions plus a model-level round trip from the literal wire body. Sibling flexible types checked and noted there.

PR #664, opened 2026-08-04
Schedule entries: a bare date must be creatable, not only replaceable

### LINKED ISSUE #634: CreateScheduleEntry cannot express an all-day bare date, so create and replace disagree
Found while implementing the ScheduleEntries triad (#632), reported rather than absorbed.

`CreateScheduleEntry`'s `StartsAt` is a plain Go `time.Time`, so it **cannot express a bare date** — the form BC3 accepts for an all-day entry.

`ReplaceScheduleEntry` can. So the two operations disagree about what an all-day entry looks like: a caller can *replace* an entry into a bare-date all-day state that they could never have *created* directly.

That asymmetry is the kind of thing a consumer discovers only after writing both paths and finding one of them impossible.

## Related but distinct

#633 covers `types.FlexibleTime` round-tripping a bare date as a midnight timestamp. This issue is narrower: the create operation's field **type** cannot carry the bare-date form at all, regardless of how `FlexibleTime` marshals.

Fixing #633 alone will not close this; fixing this alone leaves the round-trip corruption in place. They want to be resolved together, or at least with each aware of the other.

## What a fix needs

Model `CreateScheduleEntry.StartsAt`/`EndsAt` so a bare date is expressible, matching whatever `ReplaceScheduleEntry` settles on — the two must agree.

Verify against bc3 at the pinned provenance revision what the create endpoint actually accepts before modelling, rather than inferring symmetry from the replace path. `doc/api` is **not** authoritative in the affirmative direction here; check `routes.rb`, the controller's permitted params, and `test/api`.

jeremy:
Closes #634.

## The asymmetry, in source

An all-day schedule entry's bounds are **bare dates** on the wire. Go could send one on replace and not on create.

**Create** — `go/pkg/basecamp/schedules.go` (at `origin/main`, `e0431a722`):

```go
startsAt, parseErr := time.Parse(time.RFC3339, req.StartsAt)
if parseErr != nil {
    err = ErrUsage("schedule entry starts_at must be in RFC3339 format (e.g., 2024-01-15T09:00:00Z)")
    return nil, err
}
```

…feeding `generated.CreateScheduleEntryRequestContent`, where `StartsAt time.Time` / `EndsAt time.Time`.

**Replace** — the same file, `replaceScheduleEntry`, already names the problem and works around it:

> The body is a hand-marshaled map rather than the generated request struct: […] it types `starts_at`/`ends_at` as `time.Time`, which cannot express the bare date BC3 renders for an all-day entry at all.

So `ReplaceScheduleEntryRequest.StartsAt *string` goes on the wire verbatim, and `CreateScheduleEntryRequest.StartsAt string` could not. A caller could REPLACE an entry into an all-day state they could never CREATE.

**Which SDKs**: Go only. `starts_at` is `ISO8601Timestamp` in `spec/basecamp.smithy` — `string` — and the OpenAPI document says `"type": "string"` for both request bodies. Python `str`, Swift `String`, TypeScript `string`, Kotlin `String`, Ruby a plain kwarg. Only Go narrowed it, via the blanket `_at` pass in `scripts/enhance-openapi-go-types.sh`, whose fifth pass stated the premise this PR disproves:

```
# Only the response schema needs this; request schemas keep time.Time since we always send RFC3339.
```

**BC3, at the pinned provenance revision `4e34dc83`** (not `doc/api`, as the issue asked):

- `Schedules::Entries::BaseController#base_schedule_entry_params` permits `:starts_at, :ends_at`, and it is the **one** permit list behind both `new_schedule_entry_params` and `update_schedule_entry_params`. Whatever update accepts, create accepts.
- `Schedule::Entry` does no format-specific parsing of either bound. ActiveRecord casts the string; `skip_time_zone_conversion_for_attributes = [:starts_at, :ends_at]` leaves it alone; `before_save :ensure_all_day_times_are_midnight_utc` → `set_times_to_midnight_utc` assigns `starts_at.to_date`.

## The fix, and where it lives

**Not the spec** for the type — the spec was already correct and symmetric. The defect was in the **Go generation overlay**, so that is where the type fix lives:

- `scripts/enhance-openapi-go-types.sh` gains a fifth-a1 pass stripping `x-go-type`/`x-go-type-import` from `starts_at`/`ends_at` on `CreateScheduleEntryRequestContent` **and** `ReplaceScheduleEntryRequestContent`, leaving the plain `string` the OpenAPI document already declares. Both, because the generated layer had the identical latent defect on replace — the wrapper just bypassed it.
- `go/pkg/basecamp/schedules.go`: `CreateEntry` drops the `time.RFC3339` parse and threads the caller's string through verbatim.

**The spec does change, for documentation.** `CreateScheduleEntryInput.starts_at`/`ends_at` and `ReplaceScheduleEntryInput.starts_at`/`ends_at` now state the date-or-timestamp rule and cite the shared permit list. That is not cosmetic: it displaced a generated fallback that was actively telling callers the wrong thing —

```diff
-      # @param starts_at [String] starts at (RFC3339 (e.g., 2024-12-15T09:00:00Z))
+      # @param starts_at [String] The entry's start, as a bare date ("2026-06-01") for an all-day entry…
```

…in Ruby, and the same in TypeScript. `SPEC.md` §schedule-entry write surface gains the create-side sentence.

**`types.FlexibleTime` is deliberately not used here.** It decodes a bare date but marshals it back as a midnight timestamp — the same loss one layer down. That is #633, which stays deferred and is not absorbed.

## Red proof

The test asserts the **bytes on the wire**, not a type and not an error class: a bare date survives the create path, and create and replace put the identical value there.

`go/pkg/basecamp/schedules_test.go` → `TestSchedulesService_CreateEntryAllDayBareDate`, plus conformance case `create-all-day-bare-date` in `conformance/tests/schedule_entries_write.json` (six runners).

**RED** — un-fixed sources restored with `git show origin/main:<path> > <path>` into the worktree (no `git stash`), test file kept:

```
=== RUN   TestSchedulesService_CreateEntryAllDayBareDate
    schedules_test.go:1795: create with bare dates failed: schedule entry starts_at must be in RFC3339 format (e.g., 2024-01-15T09:00:00Z)
--- FAIL: TestSchedulesService_CreateEntryAllDayBareDate (0.00s)
FAIL
FAIL	github.com/basecamp/basecamp-sdk/go/pkg/basecamp	0.235s
FAIL
MARKER_REAL_EXIT_RED_GO=1
```

**RED, second half** — fixed wrapper against the un-fixed generated client, proving both layers are load-bearing:

```
# github.com/basecamp/basecamp-sdk/go/pkg/basecamp [github.com/basecamp/basecamp-sdk/go/pkg/basecamp.test]
pkg/basecamp/schedules.go:871:16: cannot use req.StartsAt (variable of type string) as "time".Time value in struct literal
pkg/basecamp/schedules.go:872:16: cannot use req.EndsAt (variable of type string) as "time".Time value in struct literal
FAIL	github.com/basecamp/basecamp-sdk/go/pkg/basecamp [build failed]
FAIL
MARKER_REAL_EXIT_RED_GENONLY=1
```

**GREEN** — fix restored:

```
=== RUN   TestSchedulesService_CreateEntryAllDayBareDate
--- PASS: TestSchedulesService_CreateEntryAllDayBareDate (0.00s)
PASS
ok  	github.com/basecamp/basecamp-sdk/go/pkg/basecamp	0.230s
MARKER_REAL_EXIT_GREEN_GO=0
```

## What was run

`make generate` then `make check`, exit codes written into the log under a marker and grepped back:

| marker | value |
|---|---|
| `MARKER_REAL_EXIT_GENERATE` | `0` |
| `MARKER_REAL_EXIT_CHECK2` | `0` — `==> All checks passed` |
| `MARKER_REAL_EXIT_DOC` | `0` (`doc-constants-check`, re-run after the `SPEC.md` edit) |

The new conformance case passes in all six runners (five print per-case names, TypeScript prints a vitest aggregate).

The first `make check` failed on the lockfile tripwire with `conformance/runner/ruby/Gemfile.lock` and `conformance/runner/python/uv.lock` — both `.gitignore`d and untracked, created by the first conformance run in a brand-new worktree, not by this change. The second run was byte-identical across all 9 lockfiles.

`jq '.operations|length' behavior-model.json` → **247**, unchanged. This adds no operation.

## Breaking

**Yes**, and the label is on.

`generated.CreateScheduleEntryRequestContent` and `generated.ReplaceScheduleEntryRequestContent` now type `StartsAt`/`EndsAt` as `string` rather than `time.Time`. `pkg/generated` is a supported public surface (SPEC.md §7 documents its retry configuration as the low-level client), so a caller constructing either body directly breaks at compile time.

The `API Compatibility` CI check will not show this and its green mark proves nothing twice over: it is `# Don't fail the build, just warn`, **and** it runs `apidiff` against `./pkg/basecamp` only, which is the package that did not change.

**Migration** — for the low-level generated client:

```go
// before
body := generated.CreateScheduleEntryJSONRequestBody{
    StartsAt: mustParse(time.RFC3339, "2026-06-01T09:00:00Z"),
    EndsAt:   mustParse(time.RFC3339, "2026-06-01T10:00:00Z"),
}

// after
body := generated.CreateScheduleEntryJSONRequestBody{
    StartsAt: "2026-06-01T09:00:00Z",
    EndsAt:   "2026-06-01T10:00:00Z",
}
// and now expressible at all:
body := generated.CreateScheduleEntryJSONRequestBody{StartsAt: "2026-06-01", EndsAt: "2026-06-02"}
```

Holders of a `time.Time` call `t.Format(time.RFC3339)`.

`basecamp.CreateScheduleEntryRequest` needs no migration: `StartsAt`/`EndsAt` were already `string`. What changes is that `CreateEntry` no longer returns `ErrUsage` for a value `time.RFC3339` cannot parse — the set of accepted values widens, and a caller relying on that rejection loses it. Required-and-non-empty is still enforced.

## One thing the issue got wrong

It says `CreateScheduleEntry.StartsAt` "is a plain Go `time.Time`". The **public** `basecamp.CreateScheduleEntryRequest.StartsAt` was already `string`; it was the RFC3339 parse in `CreateEntry` plus the `time.Time` on the *generated* body that made the bare date unreachable. Same user-visible defect, but it explains why no gate caught it: every drift and parity gate reads the OpenAPI type, which said `string` and agreed with the other five SDKs. The narrowing happened downstream of everything that looks.

---
## Summary by cubic
Allow creating all-day schedule entries by sending bare-date bounds on create, matching replace. `starts_at`/`ends_at` are now treated as plain strings and passed through verbatim in Go. Closes #634.

- **Bug Fixes**
  - Go `CreateEntry`: removed RFC3339 parsing; sends `starts_at`/`ends_at` exactly as provided.
  - Generator: `scripts/enhance-openapi-go-types.sh` strips `x-go-type` for `CreateScheduleEntryRequestContent` and `ReplaceScheduleEntryRequestContent`, so `StartsAt`/`EndsAt` are `string` instead of `time.Time`.
  - Spec/docs: document the date-or-timestamp rule; updated comments in Ruby/TypeScript.
  - Tests: added conformance case and a Go test asserting bare dates reach the wire.

- **Migration**
  - Breaking for low-level `pkg/generated`: `generated.CreateScheduleEntryRequestContent` and `generated.ReplaceScheduleEntryRequestContent` now use `string` for `StartsAt`/`EndsAt`. Pass strings (e.g., `t.Format(time.RFC3339)`) and bare dates are allowed.
  - No change for `basecamp.CreateScheduleEntryRequest` callers; only the accepted values widen.

<sup>Written for commit 1226080c6cfa7d5c45c8d628f98a3d3e3c3a3179. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/basecamp/basecamp-sdk/pull/664?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jeremy:
Merging on independent verification; both bot reviewers were unavailable throughout this freeze (Codex not firing at head on any PR, Copilot's check failing repo-wide on its own infrastructure).

Two things here are worth recording beyond the fix itself.

**The issue's diagnosis was wrong in a way that explains the gate blind spot.** #634 says `CreateScheduleEntry.StartsAt` "is a plain Go `time.Time`". The public field was already `string`. What made a bare date unreachable was the RFC3339 parse inside `CreateEntry` plus `time.Time` on the *generated* body — and that narrowing was injected by the blanket `_at` pass in `scripts/enhance-openapi-go-types.sh`, downstream of everything that inspects types. The Smithy spec was correct and symmetric, and all six SDKs agreed on `string`, which is exactly why five of them were fine and why no drift or parity gate could see it. A Go-only overlay can narrow a type after every parity check has already looked.

Worth noting the `API Compatibility` check is doubly uninformative here: warn-only (test.yml:859 `# Don't fail the build, just warn`) *and* it runs `apidiff` against `./pkg/basecamp` only — the one package this PR does not change.

**The spec edit is load-bearing, not cosmetic.** The generated fallback documentation read `starts_at (RFC3339 (e.g., 2024-12-15T09:00:00Z))`, so Ruby and TypeScript callers were being told the opposite of the truth about a parameter their SDKs have always accepted a bare date for.

`types.FlexibleTime` was considered and rejected for the request side: it marshals a bare date back as a midnight timestamp, which is #633. **#633 is not absorbed and was not required** — it stays deferred.

Red proof is in two halves, which is the right shape because two layers had to change: the un-fixed wrapper rejects the bare date at the RFC3339 parse, and the fixed wrapper against the un-fixed generated body fails to compile (`cannot use req.StartsAt (variable of type string) as "time".Time`). The assertion is the wire bytes — `starts_at == "2026-06-01"` on create, and create and replace sending the identical value — not merely that a type changed.

Verified against BC3 at the pinned `4e34dc83`: `Schedules::Entries::BaseController#base_schedule_entry_params` is the single permit list behind both create and update, and `Schedule::Entry` does no format-specific parsing, so the server genuinely accepts the same value set on both verbs.

Operations stay 247. Full `make check` green with the real exit code written to a marker and grepped back.

PR #665, opened 2026-08-04
ci: exclude java/class-name-matches-super-class from CodeQL

jeremy:
Post-merge residue from #391 (java-kotlin CodeQL re-enable, merged `fbfc5cdd4`). The analysis itself is healthy — `manual` build mode, CLI 2.26.2, `Build (Kotlin)` compiling `:basecamp-sdk:compileKotlinJvm`, `database finalize` succeeding, alerts landing on real `commonMain` files. What's left is a false-positive class.

## The alerts

Six Kotlin composite services trip `java/class-name-matches-super-class` (`note` severity, no security severity) for sharing a name with the generated service they wrap:

| Alert | File |
|---|---|
| 159 | `services/DocumentsService.kt` |
| 158 | `services/TodolistsService.kt` |
| 121 | `services/UploadsService.kt` |
| 120 | `services/TodosService.kt` |
| 119 | `services/CardsService.kt` |
| 162 (already dismissed) | `services/SchedulesService.kt` |

That shared name *is* the SPEC.md §18 hand-written-composite convention — the query flags the convention itself, so it recurs for every future composite. One `query-filters` exclusion retires the class.

## Scoping

`query-filters` is global, but the id is Java-specific and the repo's only `.java` sources (`spec/smithy-bare-arrays/`) sit outside the config's `paths:` allowlist *and* outside the java-kotlin build (`./gradlew :basecamp-sdk:compileKotlinJvm`), so they never enter a database. No other language's coverage changes.

Rejected alternatives: `paths-ignore` and the workflow's SARIF filter both target generated code, and these files are hand-written; narrowing `queries:` off `security-and-quality` would change coverage for all six languages.

## Deliberately not excluded

`java/local-variable-is-never-read` stays active. Its one open alert (118, `oauth/Device.kt` reported at line 1, message naming `tmp0_other_with_cast`) is a Kotlin compiler-synthesized `equals` temporary, not source code — dismissing that instance individually keeps the query catching genuine unread locals in hand-written Kotlin.

## Verification

This diff touches `.github/**`, which the `config` paths filter promotes to a full-language matrix, so this PR's own CodeQL run exercises the change:

- [ ] `Analyze (java-kotlin)` succeeds; `Filter generated-code alerts from SARIF` reports **1** finding rather than 7
- [ ] The other five `Analyze (*)` jobs report unchanged finding counts
- [ ] `GitHub Actions audit` stays green (no workflow file changed, so no actionlint delta expected)
- [ ] After merge: `gh api 'repos/basecamp/basecamp-sdk/code-scanning/alerts?state=open&per_page=100'` shows alert 118 as the only surviving `kotlin/` source alert

Note the SARIF count above is evidence about the *filter*, not about extraction — extraction health remains `database finalize` completing without exit 32.

---
## Summary by cubic
Exclude the CodeQL rule `java/class-name-matches-super-class` to stop false-positive notes on Kotlin composite services that intentionally share names with the generated services they wrap. The query is Java-only; our `.java` sources are outside CodeQL `paths` and the java-kotlin build so coverage is unchanged, and we noted that if either condition changes this query would be suppressed there too; other rules like `java/local-variable-is-never-read` remain active.

<sup>Written for commit 129b55decc924a8bcfca4a52398b06065dc38fbe. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/basecamp/basecamp-sdk/pull/665?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -28,3 +28,11 @@ paths-ignore:
 query-filters:
   - exclude:
       kind: [diagnostic, metric]
+  # Kotlin composites deliberately share a name with the generated service they
+  # extend (SPEC.md §18 "Hand-Written Composite Methods"). This Java naming query
+  # flags the convention itself, so it fires for every current and future
+  # composite. The filter is global, but the id is Java-specific and the only
+  # .java sources here (spec/smithy-bare-arrays/) are outside `paths:` above and
+  # outside the java-kotlin build, so no other analysis loses coverage.
+  - exclude:
+      id: java/class-name-matches-super-class
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -33,6 +33,8 @@ query-filters:
   # flags the convention itself, so it fires for every current and future
   # composite. The filter is global, but the id is Java-specific and the only
   # .java sources here (spec/smithy-bare-arrays/) are outside `paths:` above and
-  # outside the java-kotlin build, so no other analysis loses coverage.
+  # outside the java-kotlin build, so no other analysis loses coverage. That
+  # holds only while those two conditions do: adding a Java tree to `paths:` or
+  # to the build would silently suppress this query there too.
   - exclude:
       id: java/class-name-matches-super-class


jeremy:
## Verification results

**Copilot errored twice** (`Copilot encountered an error and was unable to review this pull request`), so per repo convention Codex stood in as the substitute review. It approved, independently confirming the `exclude.id` form, the `paths:`/build scoping (including that `kotlin/settings.gradle.kts` doesn't include the spec project), and that last-matching-filter precedence can't re-enable anything here since both entries are excludes with no later `include`. Its one caveat — the exclusion is global, so the scoping claim is contingent rather than structural — is now written into the config comment (`129b55dec`).

**Step 2 — the exclusion works, and the delta is attributable.** A raw SARIF count would have been ambiguous: the PR run does diff-informed analysis (`Computing PR diff ranges... Persisted 1 diff range(s) across 1 file(s)`) while the main baseline does not (`Not performing diff-informed analysis because we are not analyzing a pull request`). Diff-informed suppression drops *results*; a `query-filters` exclusion drops the *query*. The interpreted-query roster separates them:

| | baseline (main `a3174cf3e`, run `30941981938`) | PR `7ca2eb391` |
|---|---|---|
| java-kotlin queries interpreted | 167, incl. `java/class-name-matches-super-class` | 166, that id **absent** |
| `java/local-variable-is-never-read` | interpreted | **still interpreted** |
| `java.sarif` | 7 findings | **1 finding** |

The surviving finding is the `UnreadLocal` one on unchanged `Device.kt` — which also proves diff-informed analysis is not what removed the other six.

**Step 3 — no collateral coverage loss.** Interpreted-query sets are identical between baseline and PR for every other language: go 26, javascript 138, ruby 19, swift 4, all `diff`-clean.

Raw finding counts on the other languages did move (go 6→0, javascript 5→4, swift 7→0, ruby 3→3), but all five PR jobs log `Computing PR diff ranges`, so that is uniform diff-informed suppression of security queries on a PR touching one YAML file — not this change. Worth flagging because the obvious check ("finding counts unchanged") gives a false alarm here.

**Step 1** — `GitHub Actions audit` green; 50 checks total, all green except the two flaky Copilot attempts and cubic's `neutral` (no findings).

**Extraction health** (the signal that actually matters, independent of alert counts): CLI 2.26.2, `> Task :basecamp-sdk:compileKotlinJvm`, `BUILD SUCCESSFUL in 1m 10s`, `database finalize` with no exit 32.

Alert 118 has been dismissed as a false positive with the query left enabled.

Issue #667, opened 2026-08-04
v0.13.0 release freeze — main is frozen at 931c36ad7


jeremy:
`main` is frozen at **`931c36ad7`** pending the v0.13.0 tag. Please hold non-critical merges until the tag lands; a regression fix is of course still welcome, but it re-opens the frozen proof below.

## What is in the release

**55 merged PRs since v0.12.0, 15 of them breaking.** Derive it yourself rather than trusting the figure — a hand-carried count is how these go wrong, and the first version of this one was wrong:

```bash
epoch=$(git log -1 --format=%ct v0.12.0)
gh pr list --repo basecamp/basecamp-sdk --state merged --limit 300 \
  --json number,mergedAt --jq "[.[]|select((.mergedAt|fromdateiso8601) > $epoch)]|length"
```

Use `%ct` and `fromdateiso8601`, not `%cI` and a string `>`. `%cI` emits an offset-formatted timestamp while GitHub's `mergedAt` is `Z`-formatted, and jq's `>` on strings is lexicographic — that comparison silently over-counts by 12.

Landed in the freeze itself: #648, #652, #656, #657, #658, #660, #664, #642.

`check-targets` went **41 → 43** (#648 added `smithy-mapper-test`, #652 added `check-projected-examples`). Operations stay at **247**.

Upgrade notes live in [`MIGRATING.md`](MIGRATING.md): 61 breaks survive a clean build — 55 with no signal at all, 6 that compile and then fail at runtime.

## Held deliberately, both landing immediately after the tag

- **#645** (event-feed conformance) — another program's deliverable with open review items, and it adds gates to the very `check-targets:` line the frozen proof is measured against. Merging it would invalidate the proof.
- **#610** (dependabot actions bump) — carries `astral-sh/setup-uv` v8.3.2 → **v9.0.0**, twice, inside `release-python.yml`. A major-version debut whose first real exercise would be the publish it is performing. Its v9 changelog needs checking for input renames first: a silently-ignored input in a release workflow fails quietly.

## Filed during the freeze rather than absorbed into it

[...]

PR #676, opened 2026-08-05
Python was in no security tooling at all

jeremy:

`main` is frozen. This is prepared and held. It should land **before** the Link-header PR, because it converts "Python is unscanned" from an assertion into a measured fact.

## What was wrong

`.github/workflows/codeql.yml` was created 2026-02-28 as *"Add CodeQL analysis for all 5 SDK languages"* (`a99fdc357`, #132). The Python SDK arrived 2026-03-24 as the sixth (`e1fe52925`, #200) and was never retrofitted.

Python appears in **none** of:

- the CodeQL language matrix
- `codeql-config.yml`'s `paths:`
- `security.yml` — which runs Trivy ×3, npm audit, bundler-audit, Gradle dependency check, Swift dependency audit, `govulncheck` and `gosec`

`ruff`'s `select` was `["E","F","I","UP","B","SIM"]`, so no bandit rules either. No note, no issue, no SPEC line records this as a decision — it is an omission, not a choice.

It is not academic. The six SDKs' `Link`-header parsers split cleanly, and Python's is unflagged **only because nothing looks at it**:

| SDK | Approach | Status |
|---|---|---|
| Go / Swift / Kotlin | index scan | linear |
| TypeScript | `/<([^>]+)>/` | quadratic — alert 43 |
| Ruby | `/<([^>]+)>/` | quadratic — alert 48 |
| Python | `re.search(r"<([^>]+)>")` | quadratic — **unflagged** |

The repo already treats that header as attacker-influenced (`isSameOrigin` exists "to prevent SSRF via poisoned Link headers"), so an unscanned ReDoS on it is inconsistent with the threat model the code has already adopted. The follow-up PR fixes the parsing itself; this PR is what makes the Python instance visible.

## The three changes

**1. CodeQL.** A `python` paths-filter entry, a `PYTHON` env var, and `add python none` in *both* branches of the matrix builder. Python is buildless, so unlike Go and Kotlin it needs no build step. `python/` joins `paths:`.

Deliberately **no** Python `paths-ignore`. `python/src/basecamp/generated/` holds two *hand-written* files — `services/_base.py` and `services/_async_base.py`, the HTTP, retry and pagination plumbing that AGENTS.md exempts from the generator — i.e. exactly the code most worth scanning. CodeQL path filters have no negation, so a directory exclusion would silently drop them. The config now says this, and points at `codeql.yml`'s jq SARIF filter as the place to narrow if generated Python turns out to be noisy (that step is a `test()` regex and *can* express "generated/ but not `services/_`").

**2. Trivy.** A `trivy-python` job cloned from `trivy-ruby`, keeping all four steps of the existing shape: full-severity SARIF upload under category `trivy-python`, the HIGH/CRITICAL table gate, and `assert-lockfiles-unchanged --verify-clean`. Trivy `fs` picks up `python/uv.lock`.

**3. ruff bandit.** `"S"` added to `select`. 1935 hits; 1924 are test-fixture noise cleared by one `tests/**` ignore.

> Note: the ignore list is `["S101", "S105", "S106"]`, not the `["S101","S106"]` originally scoped — tests also carry 23 `S105` hits (fake `access_token=`, `secret=`, `TOKEN_ENDPOINT` literals), same genus as S106.

Of the 11 hits in `src/`:

- **4× S101**, all the same shape: `assert self._result is not None`, narrowing a `dict | None` for mypy immediately after an explicit `raise` that already enforces the real invariant. `_result` is assigned immediately before `_completed = True` on every path, so the assert is pure narrowing. But `python -O` strips `assert`, so instead of suppressing them the None check folds into the raise that was already there — same behaviour, now actually enforced under `-O`, no `noqa`, and mypy still narrows.
- **5× S105 / 2× S311** — `TOKEN_URL` (a public endpoint), `token_type = "Bearer"` (the RFC 6750 scheme name), and `random.random()` for retry-backoff jitter. False positives; targeted `noqa` with the reason on the preceding line, since ruff parses trailing text after a noqa code unreliably.

`generated/` has zero S hits, so its per-file-ignores are unchanged.

## Out of scope, noted

`python/scripts/` is not linted by CI at all — `test.yml` lints only `src/ tests/`. Same gap as root `scripts/`.

## Verification

- `ruff check src/ tests/` clean; `ruff format --check` clean (160 files)
- `mypy src/` clean over 39 files — this is the proof the raise-based narrowing works
- 1178 tests pass, 4 skipped
- `actionlint` and `zizmor` both clean on the two changed workflows
- Matrix builder extracted from the workflow and simulated across four cases: push (full matrix, python present), config-touch PR (full matrix), python-only PR (python alone), ruby-only PR (python correctly absent)
- `uv.lock` untouched, so `assert-lockfiles-unchanged` stays clean

This PR touches `.github/**`, which the `config` filter promotes to a full matrix — so its own run exercises the new job. Reviewers: confirm an `Analyze (python)` job appears and succeeds, and read `python.sarif: N findings` off its filter step. Compare **interpreted-query rosters** rather than finding counts — PR runs are diff-informed and main runs are not, so counts are not comparable across the two.

---
## Summary by cubic
Adds the Python SDK to our security tooling so it’s scanned like the other languages. This enables `CodeQL`, `Trivy`, and `ruff` bandit checks, and tightens a few guards to satisfy new rules.

- New Features
  - `CodeQL`: add `python/**` to the matrix and include `python/` in `paths` (no `paths-ignore` so the hand-written `services/_base.py` files are scanned).
  - `Trivy`: new job for `python/` with SARIF upload, HIGH/CRITICAL gate, and lockfile integrity check.
  - Linting: enable `ruff` bandit (`"S"`); ignore `tests/**` for `S101/S105/S106`; add targeted `noqa` for public constants (`S105`) and backoff jitter (`S311`).
  - Code fixes: replace `assert`-based result checks with explicit raises to keep behavior under `-O` and pass `S101`.

<sup>Written for commit 162005f6ac55860e4588c34b4e90be18dcd5959a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/basecamp/basecamp-sdk/pull/676?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -8,6 +8,7 @@ paths:
   - ruby/
   - swift/
   - kotlin/
+  - python/
 
 paths-ignore:
   # Generated SDK code (fix in the generator or spec, not here).
@@ -21,6 +22,13 @@ paths-ignore:
   - ruby/lib/basecamp/generated/
   - swift/Sources/Basecamp/Generated/
   - kotlin/sdk/src/commonMain/kotlin/com/basecamp/sdk/generated/
+  # python/src/basecamp/generated/ is deliberately NOT listed. It holds two
+  # hand-written files — services/_base.py and services/_async_base.py, the HTTP,
+  # retry and pagination plumbing (AGENTS.md exempts them from the generator) —
+  # i.e. exactly the code most worth scanning. paths-ignore has no negation, so
+  # excluding the directory would silently drop them. If generated Python turns
+  # out to be noisy, narrow it in codeql.yml's SARIF filter instead: that step is
+  # a jq test() regex and can express "generated/ but not services/_".
   # Test infrastructure
   - conformance/
   - kotlin/conformance/
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -41,6 +41,8 @@ jobs:
               - 'swift/**'
             kotlin:
               - 'kotlin/**'
+            python:
+              - 'python/**'
             actions:
               - '.github/**'
             config:
@@ -56,6 +58,7 @@ jobs:
           RUBY: ${{ steps.filter.outputs.ruby }}
           SWIFT: ${{ steps.filter.outputs.swift }}
           KOTLIN: ${{ steps.filter.outputs.kotlin }}
+          PYTHON: ${{ steps.filter.outputs.python }}
           ACTIONS: ${{ steps.filter.outputs.actions }}
           CONFIG: ${{ steps.filter.outputs.config }}
         run: |
@@ -67,6 +70,7 @@ jobs:
             add go manual
             add javascript none
             add ruby none
+            add python none
             add java-kotlin manual
             swift=true
           else
@@ -74,6 +78,7 @@ jobs:
             if [ "$GO" = "true" ]; then add go manual; fi
             if [ "$TS" = "true" ]; then add javascript none; fi
             if [ "$RUBY" = "true" ]; then add ruby none; fi
+            if [ "$PYTHON" = "true" ]; then add python none; fi
             if [ "$KOTLIN" = "true" ]; then add java-kotlin manual; fi
             swift=$SWIFT
           fi
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -298,6 +298,62 @@ jobs:
         working-directory: .
         run: ./scripts/assert-lockfiles-unchanged --verify-clean
 
+  trivy-python:
+    name: Trivy (Python SDK)
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      security-events: write
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      # Full-severity SARIF for Security-tab visibility; the gate step below enforces policy.
+      # (In SARIF mode the action reports all severities unless limit-severities-for-sarif is set.)
+      - name: Run Trivy vulnerability scanner
+        uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0
+        with:
+          scan-type: 'fs'
+          scan-ref: './python'
+          ignore-unfixed: true
+          format: 'sarif'
+          output: 'trivy-python-results.sarif'
+
+      - name: Upload Trivy scan results to GitHub Security tab
+        uses: github/codeql-action/upload-sarif@7188fc363630916deb702c7fdcf4e481b751f97a # v4.37.1
+        if: always()
+        continue-on-error: true
+        with:
+          sarif_file: 'trivy-python-results.sarif'
+          category: 'trivy-python'
+
+      # Table mode passes the severity filter through, so this gates on HIGH/CRITICAL only.
+      - name: Fail on HIGH/CRITICAL vulnerabilities
+        uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0
+        env:
+          TRIVY_SKIP_DB_UPDATE: true
+        with:
+          scan-type: 'fs'
+          scan-ref: './python'
+          severity: 'HIGH,CRITICAL'
+          exit-code: '1'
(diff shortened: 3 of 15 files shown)


jeremy:
### Proving the new ruleset is actually live, not suppressed into silence

Adding `"S"` alongside a `tests/**` per-file-ignore and seven targeted `noqa`s invites a fair question: is anything still being enforced in `src/`, or did the suppressions swallow the rule set? A green `ruff check` is consistent with both.

Introduced a temporary violation in `src/basecamp/_pagination.py`:

```python
def _tmp_probe(cmd: str) -> None:
    import subprocess
    subprocess.call(cmd, shell=True)
    assert cmd
```

```
src/basecamp/_pagination.py:71:5: S602 `subprocess` call with `shell=True` identified, security issue
src/basecamp/_pagination.py:72:5: S101 Use of `assert` detected
Found 2 errors.
```

Both fire. `S101` in particular is the rule the `tests/**` ignore turns off — it is still active in `src/`, which is the half that matters. File restored byte-for-byte afterwards; `ruff check src/ tests/` clean.

### Two other checks worth recording

**Adding `python/` to `paths:` cannot leak into the other analyses.** Path filters in `codeql-config.yml` are global rather than per-language, so a stray `.js`/`.rb`/`.go` file under `python/` would silently widen the JavaScript, Ruby or Go databases. There are zero such files (`find python -type f \( -name '*.js' -o -name '*.ts' -o -name '*.rb' -o -name '*.go' \)` → 0), and the interpreted-query rosters for all six pre-existing languages are byte-identical between this PR's run and the `main` baseline.

**The assert→raise rewrite changes no tested behaviour.** The existing suites assert `pytest.raises(RuntimeError, match="edit has not completed")` in eight places — that is the `not self._completed` branch, which is preserved exactly. Nothing asserts `AssertionError`, and the `_result is None` arm is unreachable (`_result` is assigned immediately before `_completed = True` on every path). The rewrite makes that arm enforced under `python -O` instead of stripped.

jeremy:

Copilot was requested, errored (`Copilot encountered an error and was unable to review this pull request`), was re-requested once, and errored again. Codex did not fire. Zero review threads — the count reflects genuinely empty reviews, not suppressed feedback.

Per the #657 precedent, independent adversarial review stands in. It is recorded in the comment above: the live-ruleset probe (S602/S101 fire in `src/` despite the `tests/**` ignore), the zero-non-Python-files check behind adding `python/` to a global `paths:`, and the confirmation that the assert→raise rewrite changes no tested behaviour.

Two further points a reviewer should weigh rather than take from me:

**The `paths-ignore` omission is the deliberate part.** Not adding `python/src/basecamp/generated/` is what keeps `services/_base.py` and `services/_async_base.py` — hand-written HTTP, retry and pagination plumbing — under analysis. CodeQL path filters have no negation, so the directory-level exclusion that looks obvious here would silently drop exactly the code most worth scanning. The config now carries that reasoning inline so the next reader does not "fix" it.

**58 findings is a measurement, not a verdict.** The filter step reports `python.sarif: 58 findings, filtered 0 generated-code alerts` — 0 because no Python path is in the SARIF regex. Those become inspectable alerts only once this merges to `main`. That is the right moment to decide whether anything needs narrowing, and the recommendation stands that if it does, it belongs in the workflow's jq filter (which can express "generated/ but not `services/_`") rather than `paths-ignore` (which cannot).

Extraction evidence is `database finalize` completing, not the finding count — carried forward from #665. CI: 48 checks green, 0 failing.

jeremy:
@codex review

jeremy:
**Correction to the previous comment:** Codex did review this PR. It responded after that comment was posted, so "both bot reviewers unavailable" and "Codex did not fire" were wrong — the `@codex review` trigger works, I checked too early.

> Codex Review: Didn't find any major issues.
> **Reviewed commit:** `162005f6ac`

That is the current head, so it is a matched-head review rather than one against a stale commit, and it opened no review threads.

What stands from that comment: Copilot errored on both its initial and re-requested review, and the independent adversarial review — the live-ruleset probe, the paths-leak check and the exception-type verification — is unaffected. The correct summary is that one of the two bot reviewers was unavailable and the other reviewed clean.

PR #677, opened 2026-08-05
Delete five locals nothing reads

jeremy:

Independent of the other security-scanning PRs; can land any time after the freeze lifts.

## What

CodeQL's `js/unused-local-variable` and `rb/useless-assignment-to-local` — alerts 86, 87, 88, 89 and 115. All five are genuinely unread, so they are worth deleting rather than dismissing.

| File | Symbol | Note |
|---|---|---|
| `typescript/scripts/generate-services.ts` | `resolveSchema` | declared, never called |
| `typescript/scripts/generate-services.ts` | `requestInterfaceName`, `optionsInterfaceName` | destructured out of `buildMethodSignature`, unused in `generateMethod` |
| `typescript/tests/auth-strategy.test.ts` | `vi` import | no `vi.*` call in the file |
| `ruby/test/basecamp/oauth_device_test.rb` | `sleeper` | line 856 |

Two details worth stating:

- The two interface names are still **returned by** and **used inside** `buildMethodSignature`. Only the unused destructure at the call site goes.
- Every other `recording_sleeper` call site threads the value through as `sleeper:`. This one does not, and `recording_sleeper` is pure (it builds an array and a lambda), so the whole line goes rather than being renamed to `_sleeper`.

## Why lint didn't catch them

This turned out to differ by language, and neither is a gap in how the linters are invoked:

- **TypeScript.** The lint script is `oxlint src --ignore-pattern 'src/generated/**'` — it lints `src` only. Both `scripts/` and `tests/` sit outside the target, which accounts for all three TS findings. (Same shape as the root `scripts/` gap, and as `python/scripts/` noted in #676.)
- **Ruby.** `Lint/UselessAssignment` is `Enabled: false` in the shared `rubocop-37signals` house style. Running `rubocop --only Lint/UselessAssignment` *does* flag line 856; a default run reports no offenses.

So this is code the linters are configured not to look at, found by the layer that does — which is a small argument for the scanning layer earning its keep.

## Verification

- `tsc --noEmit` clean
- `npm run lint` clean
- `npx tsx scripts/generate-services.ts` re-run produces a **byte-identical** tree — 53 services, 247 operations, empty `git status` under `typescript/src/generated/`. This is the one that matters: the generator's output must not move.
- `auth-strategy.test.ts` — 9 tests pass
- `oauth_device_test.rb` — 82 runs, 294 assertions, 0 failures

---
## Summary by cubic
Remove five unused locals reported by CodeQL across TS scripts/tests and a Ruby test. No behavior change; generator output unchanged and all tests pass.

- **Refactors**
  - `typescript/scripts/generate-services.ts`: delete unused `resolveSchema`; stop destructuring unused `requestInterfaceName` and `optionsInterfaceName`.
  - `typescript/tests/auth-strategy.test.ts`: remove unused `vi` import.
  - `ruby/test/basecamp/oauth_device_test.rb`: remove unused `sleeper` assignment.

<sup>Written for commit 39822d600fe8e5ad0ee7c30d98a5aff0d2ee667c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/basecamp/basecamp-sdk/pull/677?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/ruby/test/basecamp/oauth_device_test.rb
+++ b/ruby/test/basecamp/oauth_device_test.rb
@@ -853,7 +853,6 @@ class OAuthDeviceTest < Minitest::Test
     # Adapter-only: the SSRF guard refuses middleware it cannot verify
     # redirect-free (Faraday.new's default stack includes UrlEncoded).
     connection = Faraday.new { |f| f.adapter :net_http }
-    _waits, sleeper = recording_sleeper
     started = Process.clock_gettime(Process::CLOCK_MONOTONIC)
     error = assert_raises(Basecamp::Oauth::DeviceFlowError) do
       Basecamp::Oauth::DeviceFlow.request_device_authorization(
--- a/typescript/scripts/generate-services.ts
+++ b/typescript/scripts/generate-services.ts
@@ -574,14 +574,6 @@ function resolveRef(ref: string): string {
   return ref.split("/").pop() || "";
 }
 
-function resolveSchema(schemaOrRef: Schema): Schema | undefined {
-  if (schemaOrRef.$ref) {
-    const refName = resolveRef(schemaOrRef.$ref);
-    return globalSchemas[refName];
-  }
-  return schemaOrRef;
-}
-
 function getSchemaProperties(schemaRef: string): { properties: Record<string, Schema>; required: string[] } {
   const schema = globalSchemas[schemaRef];
   if (!schema) return { properties: {}, required: [] };
@@ -1162,7 +1154,7 @@ function generateMethod(op: ParsedOperation, serviceName: string): string[] {
   const resourceName = singularize(serviceName);
 
   // Build param string and types
-  const { paramString, hasOptions, hasRequest, requestInterfaceName, optionsInterfaceName } = buildMethodSignature(op, resourceName);
+  const { paramString, hasOptions, hasRequest } = buildMethodSignature(op, resourceName);
 
   // Return type
   const returnType = buildReturnType(op, serviceName);
--- a/typescript/tests/auth-strategy.test.ts
+++ b/typescript/tests/auth-strategy.test.ts
@@ -1,7 +1,7 @@
 /**
  * AuthStrategy tests
  */
-import { describe, it, expect, vi } from "vitest";
+import { describe, it, expect } from "vitest";
 import { http, HttpResponse } from "msw";
 import { server } from "./setup.js";
 import { BearerAuth, bearerAuth } from "../src/auth-strategy.js";


jeremy:

Copilot was requested, errored (`Copilot encountered an error and was unable to review this pull request`), was re-requested once per the usual practice, and errored again. Codex did not fire. Zero review threads on this PR — there is no suppressed or unresolved feedback hiding behind that count; the reviews genuinely contain no content.

Following the #657 precedent for this situation, here is what independent adversarial review covered instead.

**The claim that matters is that the generator's output does not move.** `resolveSchema` was dead, but `requestInterfaceName` and `optionsInterfaceName` were live symbols being destructured at a call site that ignored them — a plausible place to break codegen by deleting the wrong thing. Re-ran `npx tsx scripts/generate-services.ts`: 53 services, 247 operations, and `git status` under `typescript/src/generated/` empty. Byte-identical.

**The two interface names are still used** inside `buildMethodSignature` (`:1457`, `:1483`) and still returned by it (`:1491`, `:1492`). Only the unused destructure at the call site is gone, confirmed by grep before and after.

**The Ruby local is dead at that one site only.** Every other `recording_sleeper` call threads the value through as `sleeper:`; line 856 does not. `recording_sleeper` is pure — it returns `[waits, ->(seconds){ waits << seconds }]` and touches nothing — so removing the whole line is safe rather than renaming to `_sleeper`.

**The lint-gap explanation was verified rather than assumed.** I ran `rubocop --only Lint/UselessAssignment` against the unmodified file and it flagged line 856; a default run reports no offenses. `bundle exec rubocop --show-cops Lint/UselessAssignment` shows `Enabled: false`, inherited from the `rubocop-37signals` gem. On the TypeScript side the lint script is literally `oxlint src --ignore-pattern 'src/generated/**'`, so `scripts/` and `tests/` are outside the target. Two different causes, neither a misconfiguration in this repo.

Tests: `auth-strategy.test.ts` 9 passed; `oauth_device_test.rb` 82 runs / 294 assertions / 0 failures; `tsc --noEmit` and `npm run lint` clean. CI: 41 checks green, 0 failing.

jeremy:
@codex review

jeremy:
**Correction to the previous comment:** Codex did review this PR. It responded after that comment was posted, so "both bot reviewers unavailable" and "Codex did not fire" were wrong — the `@codex review` trigger works, I checked too early.

> Codex Review: Didn't find any major issues.
> **Reviewed commit:** `39822d600f`

That is the current head, so it is a matched-head review rather than one against a stale commit, and it opened no review threads.

What stands from that comment: Copilot errored on both its initial and re-requested review, and the independent adversarial review — the byte-identical generator re-run and the verified lint-gap root cause — is unaffected. The correct summary is that one of the two bot reviewers was unavailable and the other reviewed clean.

Issue #655, opened 2026-08-04
Flaky under parallel load: the caller-signal abort test races a 50ms timer against a 1000ms mocked response


jeremy:
Found while running `make check` under parallel load during #652. **Unrelated to that PR** — its diff contains no TypeScript — but it is a latent red that will flake in CI.

## The test

`typescript/tests/client.test.ts:567` — *"propagates a caller-supplied signal through the combined signal"*.

It races two timers:

```js
http.get(`${BASE_URL}/projects.json`, async () => {
  await new Promise(resolve => setTimeout(resolve, 1000));   // mocked response at 1000ms
  return HttpResponse.json([]);
})
...
const abortTimer = setTimeout(() => controller.abort(), 50); // abort at 50ms
```

then asserts both:

```js
await expect(...).rejects.toMatchObject({ name: "AbortError" });
expect(Date.now() - startedAt).toBeLessThan(900);
```

## The failure

Under `make check`'s parallel load — six SDK suites running at once — the 50 ms abort timer can slip past the 1000 ms mocked response. The request then resolves normally, no `AbortError` is thrown, and the test fails. Observed once during a #652 verification run; cleared 3/3 in isolation immediately afterward, and the full `ts-test` target then passed at 82 files / 1400 tests.

So it is timing-sensitive, not broken. Which is the worst kind of red: it will fail rarely, on an unrelated PR, and cost someone an hour deciding whether they caused it.

## The fix is smaller than it looks

**The wall-clock assertion is redundant with the assertion that actually proves the property.** The test's own comment says so:

> A caller abort surfaces as `AbortError`, distinct from the timeout's `TimeoutError` — so this also proves it was the caller's signal, not the (30s) timeout, that won.

Exactly. `name: "AbortError"` already discriminates the caller's signal from the 30 s request timeout. `Date.now() - startedAt < 900` adds no discriminating power and contributes all of the load sensitivity. Dropping it removes half the flake surface for free.

For the remaining half — the 50 ms vs 1000 ms race — options, roughly in order of preference:

1. **Abort deterministically instead of on a timer.** Call `controller.abort()` after the request is in flight (e.g. from within the MSW handler, or on a resolved microtask) so no wall-clock margin is involved at all.
2. **Fake timers**, so scheduling is not subject to machine load.
3. **Widen the margin** (abort at 50 ms, respond at 10 s). Cheapest, still probabilistic — a mitigation, not a fix.

[...]

Issue #573, opened 2026-08-01
Kotlin conformance runner accepts an over-fetch on auto-paginating fixtures


jeremy:
Split out of #558, where Codex found the same defect in the Swift runner.

`kotlin/conformance/src/main/kotlin/com/basecamp/sdk/conformance/Main.kt` evaluates `requestCount` as a lower bound whenever the fixture carries a `Link` `rel="next"` header:

```kotlin
if (autoPaginates) {
    if (requestCount < expected) {
        return TestResult(false, "Expected >= $expected requests (SDK auto-paginates), got $requestCount")
    }
} else if (requestCount != expected) { ... }
```

That is backwards for the fixtures it covers. Three of them queue MORE pages than they expect requests, because stopping early is the behavior under test:

| Fixture | Queued pages | Expected requests |
|---|---|---|
| Pagination stops at maxPages safety cap | 3 | 2 |
| maxItems caps results across pages | 3 | 2 |
| Auto-pagination follows Link headers across multiple pages | 3 | 3 |

An SDK that ignored the cap and walked all three passes `3 >= 2`. The last one is the exposed case: its only assertions are `requestCount` and `noError`, so an over-fetch has nothing else to catch it. The other two happen to carry a `responseMeta` truncated assertion that fires instead — coverage by luck, not by design.

The relaxation was protecting exactly one fixture, "List operation returns first page with Link header", and that one is already excluded earlier by the `link-header` tag branch, so nothing still reaching the evaluator needs it.

**Fix:** compare exactly, and drop the now-unused `autoPaginates` parameter from the evaluator. Keep the MockEngine's auto-pagination tolerance — answering an over-walk with a terminal empty page rather than an error is what lets the tightened assertion report a clean count mismatch.

Swift took this in #558 with the suite unchanged (136 passed, 0 failed, 1 skipped): it already made exactly the expected number of requests everywhere, so the lower bound was pure false-green surface. Kotlin is expected to behave the same, but that should be confirmed by running it rather than assumed.

Worth checking whether TypeScript, Go, Ruby and Python carry the same shape while in there.

PR #678, opened 2026-08-05
Link-header parsing: all six SDKs were wrong, in two different directions

jeremy:

Land **after** #676 (Python security tooling). That PR is what makes the Python instance of this bug visible in the first place; merging this first would quietly fix an alert that was never allowed to exist.

Closes the ReDoS half of alerts **43** (`js/polynomial-redos`) and **48** (`rb/polynomial-redos`), plus the unflagged Python equivalent.

## The framing changed during investigation

The obvious read — "make the three regex SDKs copy the three index-scan SDKs" — is **wrong**, because the two families are not equivalent:

- The **regex** `<([^>]+)>` anchors on the first `<` and captures to the first `>` **after it**.
- **Go/Swift/Kotlin** call `indexOf('<')` and `indexOf('>')` **independently from position 0**. When a `>` precedes the `<`, `end < start`, the guard fails, and extraction silently yields nothing for that part.

So all six were wrong, in different directions: three are quadratic, three mis-parse. One change repairs both — search for `>` from *after* the `<`:

| SDK | Was | Defect |
|---|---|---|
| Go | `strings.Index` ×2 from 0 | mis-parse |
| Swift | `firstIndex(of:)` ×2 | mis-parse |
| Kotlin | `indexOf` ×2 from 0 | mis-parse |
| TypeScript | `/<([^>]+)>/` | quadratic (alert 43) + early return |
| Ruby | `/<([^>]+)>/` | flagged (alert 48), see below |
| Python | `re.search(r"<([^>]+)>")` | quadratic, **unflagged** |

This is linear **and** regex-equivalent, so it is one shape in all six.

### One deviation from the plan, and it matters

The scoped one-liner was `end = indexOf('>', start + 1)`. That is *not* exactly regex-equivalent: `[^>]+` requires at least one character, so `<>` is **not** a match and the regex moves on to the next `<`. The bare one-liner would return `""` instead. On a header like `<> <https://…/page2>; rel="next"` the two disagree.

Since parity with the old output is the stated bar, the implementations skip an empty `<>` and keep scanning. It stays linear — the scan that found `<>` was O(1), so the loop only ever repeats after work it did not do. There is a dedicated test for this case in all six SDKs.

## Measured, not assumed

Header of N `<` with no reachable `>`:

| n | TypeScript | Python | Ruby |
|---|---|---|---|
| 2,000 | 7.1ms | 22.4ms | 0.1ms |
| 8,000 | 106.1ms | 275.7ms | 0.5ms |
| 32,000 | 1620.4ms | 5177.1ms | 2.1ms |

**TypeScript and Python are quadratic** — 4× the input, ~16× the work. After the fix, n=3,200,000 takes 0.8ms and 2.6ms respectively.

**Ruby is not.** Measured linear out to 3.2M characters (290ms), including with a `>` present but unreachable, which defeats the literal-prescan shortcut Onigmo would otherwise use to bail early. Alert 48 is a pattern-level finding the engine does not honour. The Ruby change here is consistency and correctness, **not** a remediation, and none of its new tests fail against the old code — they are characterization tests. Worth stating plainly, because it would be easy to let Ruby ride on TypeScript's and Python's evidence.

## Deliberate behaviour change: TypeScript

`parseNextLink` did `return match?.[1] ?? null` on the **first** part containing `rel="next"`, so one malformed segment short-circuited the whole header to `null`. The other five fall through and keep scanning. TypeScript now matches the majority:

```
parseNextLink('<malformed; rel="next", <https://api.example.com/page2>; rel="next"')
  before: null
  after:  "https://api.example.com/page2"
```

### That change also fixes truncation reporting

Found while self-reviewing, and worth stating because it reaches further than the next-page URL. `base.ts` uses `parseNextLink(...) !== null` as the `truncated` / `hasMore` signal in five places (`:76`, `:402`, `:492`, `:569`, `:576`), and `client.go:579` does the same with `!= ""`.

So under the old TypeScript short-circuit, a header whose *first* `rel="next"` segment was malformed reported **no further pages** even when a valid `rel="next"` followed. Callers saw `truncated: false` on a genuinely truncated list. Aligning TypeScript with the other five corrects that too, not just the returned URL.

## Tests

TypeScript and Ruby had **no** dedicated parser unit test — only indirect coverage through pagination flows, always with well-formed headers. No SDK had any adversarial-input test. The Link header is attacker-influenced (`isSameOrigin` exists to stop SSRF through a poisoned one), so this was a real gap.

New `typescript/tests/pagination-utils.test.ts` and `ruby/test/basecamp/pagination_link_test.rb` (private method, called via `send` rather than widened for a test); existing Go/Kotlin/Swift/Python matrices extended. The same six adversarial cases everywhere: `<` with no `>`, `>` before `<`, a URL containing a raw `>`, multiple `<...>` pairs in one part, an empty `<>`, and a malformed part followed by a valid one. The middle cases assert **parity with the old output**.

The pathological case asserts behaviour and completion, **not** wall-clock time — this suite already has timing flakiness (#655) and a duration bound would add more.

> Note on the scoped pathological input: it must contain `rel="next"`, otherwise the parser never reaches the extraction step and the case proves nothing.

### Red proofs

Against the un-fixed code, via `git show origin/main:<path>` file swaps (never `git stash` — the stack is shared across worktrees):

- **Go** — 2 of the new cases fail: `>x<…>` yields `""` instead of the URL, and `<> <…>` yields `""`.
- **TypeScript** — 2 fail: the short-circuit case, and the pathological one at **7403ms**, which blew vitest's default 5s timeout. Suite total 7.41s un-fixed vs **6ms** fixed.
- **Python** — the pathological case takes **14.05s** un-fixed vs **under 5ms** fixed (pytest has no default timeout, so it passes either way — this one is a timing demonstration, not a failure).
- **Ruby** — all pass un-fixed, as explained above.

## Verification

- Go `go test ./...` clean; `go vet` clean; `golangci-lint` 0 issues
- TypeScript 83 files / 1411 tests pass; `tsc --noEmit` clean; `oxlint` clean
- Ruby 1382 runs, 30438 assertions, 0 failures; `rubocop` clean
- Python 1185 passed / 4 skipped; `ruff`, `ruff format`, `mypy` all clean
- Swift `PaginationTests` 38 tests, 0 failures
- Kotlin 43 tests, 0 failures — confirmed against the jvmTest **XML report**, with all seven new cases present by name, since a filter matching nothing also reports `BUILD SUCCESSFUL`

---
## Summary by cubic
Fixes Link header parsing across Go, Swift, Kotlin, TypeScript, Ruby, and Python with a linear, leftmost `<…>` scan, and adds safe pagination caps to prevent unbounded loops. Python now rejects non‑integer or infinite `max_pages`; SPEC and tests lock in the algorithm and exact error messages.

- **Bug Fixes**
  - Shared `extractAngleBracketed` parser: search `>` from after `<`, skip empty `<>`, stay O(n); TypeScript/Python replace the quadratic regex; TypeScript keeps scanning after a bad `rel="next"` part; Ruby parsing tolerates malformed UTF‑8; Go/Swift/Kotlin fix the `>`‑before‑`<` mis-parse.
  - Page-cap validation: TypeScript, Kotlin, Swift validate at construction; TypeScript also guards helpers and `BaseService` with `Number.isSafeInteger`; Python enforces `max_pages` as a positive integer (rejects 0/negative/`inf`/fractions).

- **New Features**
  - TypeScript pagination adds a safety cap: `fetchAllPages`/`paginateAll` accept `maxPages` (default `DEFAULT_MAX_PAGES`), validate eagerly, and break before reading the next Link.

<sup>Written for commit 26abdc9a64b5e39467de4baa35ad77063bd99983. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/basecamp/basecamp-sdk/pull/678?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jeremy:
### Differential verification of the regex-equivalence claim

The PR asserts the new scan is *exactly* the leftmost-match semantics of `<([^>]+)>`. That claim is load-bearing — it is why the middle test cases assert parity with the old output — so I checked it empirically rather than by argument.

Differential test: 300,000 randomized inputs over an alphabet seeded with the adversarial characters (`<`, `>`, `,`, `;`, `"`, spaces) plus whole tokens (`rel="next"`, `<>`, `<a>`, a URL), comparing the shipped `parse_next_link` against a verbatim copy of the pre-change regex implementation.

```
cases=300000 mismatches=0
```

**And the test can fail.** Re-run with the naive `find(">", start + 1)` one-liner — the version without the empty-`<>` skip — substituted in:

```
cases=300000 mismatches=10
  input='<>ta"rrrel="next"<a>x,n<>rel="next"<>'
    old='a'      new=''
  input='l<>;"c<:lbxcrel="next"ne'
    old=None     new=''
  input='.e<>xrel="next"<a>rel="next":.e/;<a>:'
    old='a'      new=''
```

So the discriminating cases are real and the differential detects them, which is the evidence that skipping empty `<>` pairs was necessary rather than fussy. `[^>]+` requires at least one character, so `<>` is not a match and the regex scans on; the one-liner would return `""` — a next-page URL that resolves back to the base and could loop.

Ported to Python because it is the fastest of the six to iterate 300k times; the algorithm is character-identical across all six SDKs.

jeremy:
@codex review

jeremy:
### Characterizing the TypeScript behaviour change, quantitatively

The short-circuit removal is the one deliberate semantic change here, so "aligns with the majority" deserved more than an assertion. Differential over 300,000 generated headers — 1–4 comma-joined segments drawn from shapes that actually reach the interesting paths (`<malformed; rel="next"`, `<>; rel="next"`, `>x<u>; rel="next"`, `<a> <b>; rel="next"`, bare `; rel="next"`, `rel="prev"`, junk) — comparing shipped `parseNextLink` against a verbatim copy of the old one:

```
 "null->value":       66068
 "value->null":           0
 "value->otherValue":     0
 "same":             233932
```

Two things worth drawing out:

**It is monotonic.** Zero `value→null` and zero `value→otherValue`. The change never loses an answer the old code produced and never changes one into something else — it only supplies an answer that was wrongly withheld. Every difference is the old code returning `null` while a valid `rel="next"` was present later in the same header.

**It is not a corner case.** 22% of these structured multi-part headers hit it. Combined with `base.ts` using `parseNextLink(...) !== null` as the `truncated`/`hasMore` signal in five places, the old behaviour meant a malformed leading segment could report a truncated list as complete.

Run separately, extraction alone (old regex *with* fallthrough vs the new scan) gives **0 mismatches over the same 300,000 cases** — so the two changes are cleanly separable: extraction is exactly equivalent, and the only semantic movement is the intended one.

#### A note on the first attempt

My initial generator reported `0` short-circuit differences, which I took as a failure of the test rather than evidence of no change: a differential that cannot produce the difference you are specifically looking for is not passing, it is not reaching the input space. Sampling random characters almost never produced two comma-separated segments both containing `rel="next"`. Generating structured segments fixed it. Flagging it because the first number looked like a clean result and was not.

jeremy:

Copilot was requested, errored (`Copilot encountered an error and was unable to review this pull request`), was re-requested once, and errored again. Codex did not fire, including after an explicit `@codex review`. Zero review threads — genuinely empty reviews, not suppressed feedback.

Per the #657 precedent, independent adversarial review stands in. The substantive results are in the two comments above (300k-case differential showing extraction is exactly regex-equivalent, plus the monotonic classification of the deliberate TypeScript change). Three further checks:

**No second parser was missed.** `rg` for `<([^>]+)>`-shaped regexes across the repo returns only comments and test strings referencing the old spelling — no production code. Each SDK has exactly one `parseNextLink`/`parse_next_link` definition, with many call sites routed through it (Go alone has 7, including `timeline.go`). A second implementation would have been the obvious way for this change to be incomplete.

**The "no `>` after this `<`" early exit is correct, not a shortcut.** All six return immediately when no `>` exists after the current `<`, rather than trying later `<` positions. That is sound because any later `<` is further right, so a `>` after it would also be a `>` after this one — the search already proved there is none. Worth stating because it looks like the kind of early exit that misses cases.

**Termination is by strict advance.** The cursor always moves to `start + 1`, and the empty-`<>` branch is the only one that loops, so each iteration consumes at least one `<`. Swift walks `String.Index` rather than integers, where the same argument holds and slicing from `endIndex` yields an empty slice instead of trapping.

The Ruby caveat stands and is the one thing I would not want lost in the noise: Onigmo does not exhibit the quadratic behaviour CodeQL flags in alert 48, measured linear to 3.2M characters. Ruby's tests here are characterization tests and pass against the un-fixed code. Only TypeScript and Python had a measured ReDoS.

CI: 44 checks green, 0 failing. `make check` green end-to-end on a Linux box (`REAL_EXIT=0`, lockfiles byte-identical across 9 files); that host skips `swift-check`, `conformance-runner-tests-swift` and `conformance-swift`, which are covered by the local Swift suite (38 tests) and this PR's `Swift Tests` and `Analyze (swift)` jobs.

jeremy:
**Correction to the previous comment:** Codex did review this PR. It responded shortly after that comment was posted, so "Codex did not fire, including after an explicit `@codex review`" was wrong — the trigger worked, I just checked too early.

> Codex Review: Didn't find any major issues. Bravo.
> **Reviewed commit:** `dff1fcdd0b`

That is the current head, so the review is at matched head rather than a stale commit, and it opened no review threads.

The rest of that comment stands: Copilot errored on both attempts, and the independent adversarial review above is unaffected. But the framing "both bot reviewers unavailable" no longer applies to this PR — one of the two reviewed it clean.

jeremy:
Rebased onto `b0567a77e` (post-#676) and pushed `dff1fcdd0` → `7b574d99e`. Review at the head, not at the original commit — the description above predates four defects found in adversarial review.

Everything the description claims checks out, including the **Swift and Kotlin red proofs it never showed**: `swift test --filter PaginationTests` → 2 failures against old code, `:basecamp-sdk:jvmTest` → `43 tests completed, 2 failed`, with all seven new cases confirmed by name in the XML report. Regex-equivalence was proven by *executing* the verbatim function bodies in all six languages over exhaustive `{<,>,a,b}^≤8` plus 400k random strings — 0 mismatches.

Four defects, three introduced by this PR:

**1. Ruby's new scan is quadratic — in the one SDK this PR admits is not a remediation.** `String#index(str, offset)` takes a *character* offset; on a non-`CR_7BIT` string Ruby walks from the start to convert it, and the skip loop advances the cursor once per `<>`.

| `"é" + "<>"*n` | new scan | `byteindex` | old regex |
|---|---|---|---|
| 10,000 | 70.5 ms | 2.2 ms | 0.3 ms |
| 40,000 | 1,494.9 ms | 8.0 ms | 1.2 ms |
| 160,000 | **20,322.8 ms** | 33.6 ms | 5.0 ms |

So Ruby traded a theoretical quadratic Onigmo never realizes for a real one it does. `byteindex`/`byteslice` are O(1) and exactly parity-preserving (0 mismatches over 200k random inputs). The comment claiming linearity "by construction" was true of the other five and false of this one.

**2. No test could have caught that.** The pathological case is `"<" * 50_000`, which under the new algorithm returns after *one* `>`-scan and never executes the skip branch — a guard against the old regex, not a characterisation of the new loop. All six gain a `("<>" * n)` case. Ruby's is gated by `Timeout.timeout(5)`: 0.16s fixed, expires broken.

**3. `fetchAllPages`/`paginateAll` were unbounded `while (true)`** — the only two pagination loops in any of the six SDKs without a cap, both exported public API. A Link header naming its own page makes their termination condition never true. Now capped by an optional `maxPages`. The break sits *before* reading the Link header, not merely in the `for` condition: these loops consume at the head and fetch at the tail, so bounding only the condition still issues a final request that is never parsed — to a URL from an attacker-influenceable header. Tests assert exact fetch counts, because the off-by-one terminates too.

**4. The Swift conformance runner holds a seventh Link parser, and it had the pre-PR defect.** `nextLinkTarget` required `hasPrefix("<") && hasSuffix(">")`, so it read `>x</projects.json?page=2>` as no next link. That helper decides which requests the LINK/PATH invariants govern, so a stricter parser there doesn't fail to notice a bug — it *invents* one, failing a correct SDK on a link it followed exactly as asked.

Also: `SPEC.md`'s `[conformance]`-tagged algorithm still prescribed the short-circuit this PR removes, and ships inside the tagged tree — all six would have released contradicting their own normative text. And the behaviour had zero cross-SDK enforcement; three fixtures now cover it, tagged `auto-follow` **not** `link-header`, since that tag is a runner directive from #573 that suppresses `requestCount` and makes Swift skip the case — the first revision was a false green in four of six runners. Both following fixtures carry a `requestPath` assertion because old Kotlin returned `""` for `<>`, which resolved to `/` and satisfied `requestCount: 2` by way of the bug.

Local: full conformance `REAL_EXIT=0` across all six runners with each new case confirmed by name; TypeScript 83 files / 1422 tests; `tsc --noEmit` and `oxlint` clean.

jeremy:
## Signoff record — head `645dd7415`

Four independent adversarial rounds. Each found real defects, so the recursion was not idle.

| Round | Target | Outcome |
|---|---|---|
| 1 — parity/tests | `dff1fcdd0` | Ruby's new scan was **quadratic** on non-ASCII (`String#index` char offsets) |
| 2 — blast radius | `dff1fcdd0` | `SPEC.md` contradicted the fix; `fetchAllPages`/`paginateAll` unbounded |
| 3 — Codex | `7b574d99e` | `byteindex` **crashed** on malformed UTF-8; cap accepted `Infinity`/`2.5`/`0` |
| 4 — Codex | `c08e5ffa2` | `Number.isInteger` admits **`Number.MAX_VALUE`** — unbounded again |

Round 4's HIGH is fixed in `645dd7415`: `Number.isSafeInteger`. `Number.isInteger(Number.MAX_VALUE)` is `true`, the counter then reaches `2 ** 53` where `page++` stops advancing, and `page === maxPages` is never reached — the same runaway `Infinity` produces, through the door the previous commit had just closed. Red proof: 6 failures on restoring `isInteger`, restored byte-identically (sha256).

### Swift precondition — probe evidence

Swift's rejected path is **not** provable in-process: `precondition` traps, XCTest has no death-test facility, and neither of the SDK's pre-existing preconditions is covered either. Its in-suite tests assert only the accepted side and say so in-code. Verified out of process instead, against a rebuilt module:

```
$ swiftc -I .build/debug/Modules -I .build/debug -L .build/debug -lBasecamp probe.swift -o probe
$ for v in 25 1 0 -3; do probe $v; echo "exit=$?"; done
maxPages=25 -> exit=0    CONSTRUCTED maxPages=25
maxPages=1  -> exit=0    CONSTRUCTED maxPages=1
maxPages=0  -> exit=133  BasecampConfig.swift:62: Precondition failed: maxPages must be positive, got: 0
maxPages=-3 -> exit=133  BasecampConfig.swift:62: Precondition failed: maxPages must be positive, got: -3
```

Exit 133 is SIGTRAP. Note the probe first reported `0` constructing **fine** — because the preceding red proof had rebuilt `.build` *without* the validation, and restoring source does not rebuild the artifact. Rebuild before probing.

### Round 4's three lower-severity findings — recorded, not fixed

1. **Ruby/Python conformance summaries** disagree with the other four on an enveloped `{"projects": []}` body (`retry.json`). Harness-only, and **no fixture asserts it**. My first fix was measured to be a **no-op** — the runner receives a `Basecamp::ListEnumerator`, never the `Hash` the guard tested for — so it was reverted rather than shipped unverified.
2. **Ruby non-ASCII-compatible encodings** (UTF-16LE) now disagree with the original parser. No Link header is ever UTF-16LE, and Codex confirmed no downstream bypass: URI parsing fails and `Security.same_origin?` rejects.
3. **TypeScript summary inherits the known >2⁵³ id precision loss.** Pre-existing; current pagination fixtures use small ids.

Also observable and worth knowing: `paginateAll.constructor.name` changes from `AsyncGeneratorFunction` to `Function`, and its function-level `prototype[Symbol.asyncIterator]` disappears, because eager validation required a wrapper. Iterator behaviour, return type, `.return()`/`.throw()`, laziness and the re-export are all unchanged.

### Status

50 checks green. The sole red is `copilot-pull-request-reviewer`, which has now failed **five** times with zero comments and zero review threads; its body reads *"Copilot encountered an error and was unable to review this pull request."* It is not a required check.

Issue #670, opened 2026-08-04
make check writes conformance/runner/ruby/Gemfile.lock — the bundler half of #612/#631 was never fixed


jeremy:
## What happens

On a machine where the Ruby conformance runner's `Gemfile.lock` is not already converged, `make check` **rewrites it** and then fails its own tripwire:

```
FAIL: the checks modified a dependency lockfile.

      34b8077165aa89ccc838821dfe2cfdfdc1de47ffebff7c2e4caa5a8657ef2056  ./conformance/runner/ruby/Gemfile.lock
      9fe4f5601dc2e5fb5800f1eb6e8c6e198fe966f36f01504c31c41c7fd6cf386b  ./conformance/runner/ruby/Gemfile.lock

Nothing `make check` runs may write a lockfile — a dirty tree blocks the
release flow, and what gets written depends on which npm ran it (#612).
Use `npm ci`, which installs from the lockfile without writing it back.
make: *** [Makefile:1282: check] Error 1
```

## Cause

Three sites run plain `bundle install`, which resolves and **writes back**:

- `Makefile:702` — `conformance-ruby`: `cd conformance/runner/ruby && bundle install --quiet && ruby runner.rb`
- `Makefile:710` — `conformance-ruby-replay`
- `Makefile:627` — the runner-test reachability target

`--quiet` suppresses output, not writes.

## Why it is the same bug as #612/#631, one language over

#631 stopped `make check` rewriting `typescript/package-lock.json`, and the tripwire message above still describes only the npm case and prescribes only the npm fix. The identical hazard on the bundler side was never addressed, and it is invisible for the same reason npm's was: on a machine whose lockfile already matches, nothing is written.

It also carries the same *specific* danger the message names — "what gets written depends on which npm ran it". Substitute bundler/RubyGems versions and the sentence holds verbatim.

## How it surfaced

During the v0.13.0 freeze, running the frozen `make check` on a Linux build host. Everything passed — Go, TypeScript, Ruby, Kotlin, Swift, Python, all 43 gates and the conformance suites — and the run failed at the very end on this, with a clean tracked tree and `PRE_SHA == POST_SHA`. So the release proof could not be completed on any machine whose runner lockfile had drifted, for a reason unrelated to the code being released.

## Fix

Make the conformance installs non-writing, matching what #631 did for npm:

```make
cd conformance/runner/ruby && bundle install --quiet --deployment && ruby runner.rb
```

[...]

PR #698, opened 2026-08-11
Sync all six SDK-version lockfiles on bump; record the upload-versions canary

### LINKED ISSUE #671: bump-version.sh syncs 4 lockfiles but 6 record the SDK version, so make check fails after every bump
## What happens

The documented release flow cannot complete:

```
scripts/bump-version.sh 0.13.0    # or: make bump VERSION=0.13.0
make check
```

```
FAIL: the checks modified a dependency lockfile.

      58dfda2b…  ./conformance/runner/ruby/Gemfile.lock
      628f1be3…  ./conformance/runner/python/uv.lock
      95099b6a…  ./conformance/runner/python/uv.lock
      e129fe59…  ./conformance/runner/ruby/Gemfile.lock
make: *** [check] Error 1
```

## Cause

Six lockfiles in this repo record the SDK's own version through a local path dependency. `bump-version.sh` syncs **four**:

| Lockfile | Records SDK version via | Synced by bump-version.sh |
|---|---|---|
| `typescript/package-lock.json` | own `package.json` | yes |
| `ruby/Gemfile.lock` | own gemspec | yes |
| `python/uv.lock` | own `pyproject.toml` | yes |
| `conformance/runner/typescript/package-lock.json` | `file:../../../typescript` | **yes** — explicitly, with a comment saying it "records the SDK's version" |
| `conformance/runner/ruby/Gemfile.lock` | `gem "basecamp-sdk", path: "../../../ruby"` | **NO** |
| `conformance/runner/python/uv.lock` | `basecamp-sdk = { path = "../../../python", editable = true }` | **NO** |

The Ruby runner's lockfile literally contains `basecamp-sdk (0.13.0)`. So after a bump those two are stale, the conformance targets run `bundle install` / `uv sync`, both re-resolve and **write**, and `assert-lockfiles-unchanged` correctly fails.

The script already knew this class of problem existed — it handles the conformance *TypeScript* runner for exactly this reason. The Ruby and Python runners were simply missed.

## Why it is not caught anywhere

`make release`'s preflight checks `uv lock --check` on `python/` and asserts the conformance **TypeScript** lockfile records the new version. It does not look at the conformance Ruby or Python lockfiles. So `make release` can pass its own guards while `make check` fails — the two disagree about what "in sync" means.

The two files are gitignored, so nothing shows up in `git status` either. On a machine that has run conformance before at the *old* version, the first post-bump `make check` fails; on a fresh machine it passes, because the lockfiles get created at the new version.

## Fix

Add both to `bump-version.sh`, mirroring the conformance TypeScript step:

```bash
echo "Syncing conformance Ruby runner lockfile..."
(cd conformance/runner/ruby && bundle install --quiet)

echo "Syncing conformance Python runner lockfile..."
(cd conformance/runner/python && uv lock --quiet)
```

and update the closing message, which currently says "Bumped 10 files" while touching lockfiles too.

Worth pairing with #670 (make the conformance installs non-writing, so a stale lockfile fails loudly instead of being silently re-resolved) — together they make the invariant "the checks never write a lockfile, and the bump leaves none stale" actually hold. Also worth extending `make release`'s preflight to assert all six, not the three it checks today, so its guards and `make check` agree.

## Provenance

Hit during the v0.13.0 release: the frozen `make check` passed cleanly at the pre-bump SHA, then failed immediately after the bump for this reason alone. Nothing about the released code was involved.

jeremy:
Pre-release housekeeping ahead of v0.14.0.

## Fix #671 — bump syncs 4 lockfiles, 6 record the SDK version

Six lockfiles record the SDK's own version through a path dependency;
`bump-version.sh` synced four. The two missed ones — the conformance Ruby and
Python runner lockfiles — are gitignored, so on any machine that had run
conformance at the old version the first post-bump `make check` re-resolved
them mid-check and `assert-lockfiles-unchanged` correctly failed.

- `bump-version.sh` now syncs both, mirroring the conformance TypeScript step:
  `bundle install --quiet` in `conformance/runner/ruby` (the same write
  `conformance-runner-tests-ruby` performs) and `uv lock --quiet` in
  `conformance/runner/python` (the lockfile-only subset of the target's
  `uv sync`)
- `make release`'s preflight asserts both — guarded on file existence, since a
  fresh clone legitimately lacks them; a machine that has run conformance is
  exactly the machine where a stale one breaks `make check`
- Closing message now says what the script does: 10 version files + 6 lockfiles

Verified: both syncs run clean at 0.13.0; the Ruby grep and `uv lock --check`
pass at the current version and the grep exits 1 against a wrong version;
`make -n release` renders the new guards as intended.

Not included: #670 (make the conformance installs non-writing) — separate
concern, this PR only closes the bump/preflight gap.

Closes #671

## Record the #683 live canary as done

The upload-versions absorption deferred one verification: a live-account
end-to-end canary. It ran today against production (account 2914079, Coworker
QA Sandbox vault) via a throwaway Go program with a `replace` directive on this
repo's `go/` module, and every assertion passed:

- `CreateVersion` kept the upload's id and URL; filename/byte_size/content_type
  reflect the replacement
- `ListVersions` decoded as typed `[]UploadVersion`; exactly one `Current`
  entry; newest action `blob_changed`
- Byte-compared both directions: the older version's `download_url` served the
  original bytes, `Download` served the replacement's
- `nil` description carried the create-time description forward
- Upload trashed afterward

`spec/api-gaps/upload-new-version.md` now records this as an as-of fact,
leaving only basecamp-cli#404's write command open there.

---
## Summary by cubic
Syncs all six SDK-version lockfiles on version bump and aligns preflight checks so stale conformance runner locks no longer break `make check`. Also records the upload-versions live canary as completed. Closes #671.

- **Bug Fixes**
  - Update `scripts/bump-version.sh` to also sync conformance runner lockfiles: Ruby via `bundle install --quiet`, Python via `uv lock --quiet`.
  - Extend `make release` preflight to assert those lockfiles are in sync, guarded on file existence for fresh clones.
  - Adjust closing message to reflect "Bumped 10 version files and synced 6 lockfiles".

<sup>Written for commit 4e7eac99800dec7f9eb4625f188abcb4a6f25e4a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/basecamp/basecamp-sdk/pull/698?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -238,6 +238,13 @@ endif
 		{ echo "ERROR: python/uv.lock is stale. Run 'make bump VERSION=$(VERSION)' first."; exit 1; }
 	@test "$$(jq -r '.packages["../../../typescript"].version' conformance/runner/typescript/package-lock.json)" = "$(VERSION)" || \
 		{ echo "ERROR: conformance/runner/typescript/package-lock.json records a stale SDK version. Run 'make bump VERSION=$(VERSION)' first."; exit 1; }
+	@# The conformance Ruby/Python runner lockfiles are gitignored and absent on a
+	@# fresh clone — check them only where they exist (a machine that has run
+	@# conformance), which is exactly where a stale one breaks `make check` (#671).
+	@test ! -f conformance/runner/ruby/Gemfile.lock || grep -qF 'basecamp-sdk ($(VERSION))' conformance/runner/ruby/Gemfile.lock || \
+		{ echo "ERROR: conformance/runner/ruby/Gemfile.lock records a stale SDK version. Run 'make bump VERSION=$(VERSION)' first."; exit 1; }
+	@test ! -f conformance/runner/python/uv.lock || (cd conformance/runner/python && uv lock --check) || \
+		{ echo "ERROR: conformance/runner/python/uv.lock is stale. Run 'make bump VERSION=$(VERSION)' first."; exit 1; }
 	@git diff --quiet && git diff --cached --quiet || \
 		{ echo "ERROR: Working tree has uncommitted changes. Commit first."; exit 1; }
 	@# Verify we're on main — release tags must be on the default branch
--- a/scripts/bump-version.sh
+++ b/scripts/bump-version.sh
@@ -76,4 +76,14 @@ echo "Syncing Python lockfile..."
 echo "Syncing conformance TypeScript runner lockfile..."
 (cd conformance/runner/typescript && npm install --package-lock-only --ignore-scripts --silent)
 
-echo "Done. Bumped 10 files to $VERSION."
+# Sync conformance Ruby and Python runner lockfiles (they record the SDK's
+# version via their path deps on ../../../ruby and ../../../python). Both are
+# gitignored: left stale, the conformance targets' installs re-resolve and
+# WRITE them mid-check, so assert-lockfiles-unchanged fails (#671).
+echo "Syncing conformance Ruby runner lockfile..."
+(cd conformance/runner/ruby && bundle install --quiet)
+
+echo "Syncing conformance Python runner lockfile..."
+(cd conformance/runner/python && uv lock --quiet)
+
+echo "Done. Bumped 10 version files and synced 6 lockfiles to $VERSION."
--- a/spec/api-gaps/upload-new-version.md
+++ b/spec/api-gaps/upload-new-version.md
@@ -135,5 +135,13 @@ documented and tested, and `visible_to_clients` was removed from the endpoint's
 reachable surface (it never set visibility — it only widened the notification
 audience, and could announce a client-invisible file to a project's clients).
 
-Still open: the live-account canary, and basecamp-cli#404's "upload new version"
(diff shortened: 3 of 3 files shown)


jeremy:
Merging on green CI + Copilot's clean review (3/3 files, no comments). Codex could not review this PR — its connector reports the account's code-review usage limit is reached — so the usual Codex verdict is absent for quota reasons, not because it found nothing. The change was also verified locally: both new sync steps run clean at 0.13.0, the new preflight checks pass at the current version, and the Ruby lockfile grep exits 1 against a wrong version.

PR #712, opened 2026-08-12
Build hygiene: clean never deletes tracked files, UTF-8-pinned reads, tracked runner lockfiles

### LINKED ISSUE #668: make clean deletes tracked files, and make check does not regenerate them
## What happens

```
make clean
make check
```

leaves the tree with a **deleted tracked file** and fails.

`go/Makefile:103` (reached via the root `clean` target → `go-clean` → `$(MAKE) -C go clean`):

```make
rm -f pkg/generated/*.gen.go
```

`go/pkg/generated/client.gen.go` is tracked:

```
$ git ls-files --error-unmatch go/pkg/generated/client.gen.go
go/pkg/generated/client.gen.go
$ make clean && git status --porcelain
 D go/pkg/generated/client.gen.go
```

Nothing in `check-targets` regenerates it — `go-check-generated-drift` regenerates into a temp dir and diffs, it does not restore the working file. So the natural "clean build" instinct produces a tree that is dirty in a way `git status` reports as *deletion of committed source*.

## Why it matters more than it looks

The other `*-clean` targets only remove genuinely ignored artifacts — `spec/build`, `typescript/dist`, `node_modules`, `swift/.build`, `ruby/vendor`, `python/.venv`. `go-clean` is the only one that removes committed content, so the asymmetry is invisible unless you check.

It bit a release freeze: a "cold build" step run before the frozen `make check` silently deleted the file, and the proof then ran against a tree missing tracked source. The run happened to fail first for an unrelated reason, which is the only thing that surfaced it.

Anyone reaching for `make clean` before a release check, a bisect, or a bug repro hits the same thing.

## Fix directions

Pick one deliberately, because they encode different intents:

1. **Stop deleting tracked output.** Narrow `go/Makefile`'s `clean` to build artifacts (`coverage.out`, `coverage.html`, `bench-*.txt`) and leave `pkg/generated/*.gen.go` alone. Cleanest if the generated Go client is meant to be committed and stable.
2. **Keep deleting it, and make `clean` restore or regenerate.** Have the target run the Go generator afterwards, so `make clean` ends with a tree that still builds.
3. **Split the targets.** Leave `clean` non-destructive and add an explicit `go-clean-generated` for the rare regenerate-from-scratch case, so the destructive option has to be asked for by name.

Whichever is chosen, a guard is worth adding: after `make clean`, assert `git status --porcelain` reports no deleted tracked files. That is a two-line check and it makes the invariant explicit instead of folklore.

## Related

The same freeze surfaced that `lint-actions` fails with an install hint when `zizmor` is merely off `PATH` rather than absent — a separate papercut, not filed, but in the same "clean-machine ergonomics" family.

### LINKED ISSUE #670: make check writes conformance/runner/ruby/Gemfile.lock — the bundler half of #612/#631 was never fixed

### LINKED ISSUE #702: Ruby test suite reads fixtures without pinned encoding, so make check fails under a C locale
Follow-on observation from #701, deliberately not fixed there.

#701 pinned `encoding: "UTF-8"` in the three `ruby/scripts/` generators, which
made `rb-check-drift` locale-independent. The next `make check` on the same
C-locale machine (non-interactive ssh, no `LANG`) then failed one layer down,
in `rb-test`:

```
ruby/test/basecamp/oauth_resource_discovery_test.rb:27:in 'block in <class:OAuthResourceDiscoveryTest>'
json/common.rb:355:in 'String#encode': "\xE2" on US-ASCII (Encoding::InvalidByteSequenceError)
```

`JSON.parse(File.read(...))` on a fixture, read under the locale's US-ASCII
default external encoding. There are likely more sites of the same shape behind
it — the run stops at the first.

## Why this wasn't fixed in #701

The LC_ALL=C hardening was scoped to **gates and generator scripts** — code
that must behave identically wherever it runs. The test suite has never been
part of that surface: CI's Ruby jobs run under the runners' UTF-8 locale and
have been green for years. The failing environment (a shell with no locale
exported) is nonrepresentative of every supported configuration, so the
release-verification cold check now exports `LC_ALL=C.UTF-8`, matching CI.

## What this issue is

A decision point, not a task: either the test suite under `LC_ALL=C` becomes a
supported configuration — then the fix is a sweep of unpinned
`File.read`/`JSON.parse` across `ruby/test/` (and a check of the other SDKs'
suites), plus a CI leg that actually runs it that way, per the existing
per-step `LC_ALL: C` pattern in `test.yml` — or it explicitly is not, and this
issue closes as the record of that decision. Adding pins without the CI leg
would be a control nobody exercises; adding the CI leg without deciding the
scope first would be armor before naming the failure.

jeremy:
Three build-hygiene fixes, one commit per issue.

Closes #668. Closes #702. Closes #670.

## `make clean` never deletes tracked files (#668)

`go/Makefile`'s `clean` removed `pkg/generated/*.gen.go`, deleting the tracked `client.gen.go` — the only `*-clean` target in the repo that touched a tracked file, and nothing in `make check` regenerates it. The `rm` moves to an explicit opt-in target (`make go-clean-generated` at the root, `clean-generated` in `go/`).

Red proof (scratch worktree on un-fixed main): `make clean` → `git status --porcelain` shows ` D go/pkg/generated/client.gen.go`. Post-fix: empty porcelain, and `make go-clean-generated && make -C go generate && git diff --exit-code go/pkg/generated` round-trips clean.

## Ruby reads pinned to UTF-8 regardless of locale (#702)

Under `LC_ALL=C`, unpinned `File.read` returns US-ASCII and non-ASCII fixtures raise. `ruby/test/basecamp/oauth_resource_discovery_test.rb` fires **today**: a class-body load error aborts the whole suite under the C locale. Every remaining unpinned read now carries `encoding: "UTF-8"` (per the 2287c7544 idiom): 5 test files, the conformance runner + replay runner (+ its test, found by sweep — one site beyond the issue), and two shipped-lib reads (`http.rb`, `config.rb` — JSON is UTF-8 per RFC 8259, so those are format-correctness). `Encoding.default_external` was deliberately rejected: it masks the locale instead of fixing the reads.

CI: the ruby 3.3 matrix leg now runs the suite under `LC_ALL: C` so the pins can't silently regress; the other rubies keep the normal locale.

Red proof: `LC_ALL=C bundle exec rake test` aborts at load pre-fix (exit 1); post-fix green under both locales (1411 runs), and `LC_ALL=C make conformance-ruby` passes (183).

## Runner lockfiles tracked and installs frozen (#670)

`conformance/runner/ruby/Gemfile.lock` and `conformance/runner/python/uv.lock` were gitignored while the TS runner's package-lock.json is tracked — so the local lockfile gate saw them but CI's `--verify-clean` (`git status --porcelain`) was structurally blind, and Linux runs silently rewrote them (PLATFORMS append + bundler-4 CHECKSUMS). Now:

- Both lockfiles tracked, regenerated fresh at 0.14.0 (the ignored ruby one still said 0.13.0), with `x86_64-linux`/`arm64-darwin` platforms locked in so macOS and Linux converge.
- All bundler recipes run `BUNDLE_FROZEN=true`; all uv recipes run `uv sync --locked` (`--locked` fails on missing **or stale**, unlike `--frozen`) — in the Makefile and in the two CI install steps, so a drift fails fast with a legible message instead of at the job-end gate.
- `--deployment` and `.bundle/config` deliberately avoided (vendored gems / untracked state).
- The release guard's `test ! -f` escape hatches (which existed only because the files used to be absent on fresh clones) are removed, and `scripts/assert-lockfiles-unchanged`'s remediation text now covers all four ecosystems.

Red proof: on un-fixed code, garbage written to both lockfiles left `--verify-clean` reporting "manifests match" (the blindness, live); post-fix the same garbage trips it. Byte-stability verified by sha256 before/after the three conformance targets on macOS **and** x86_64 Linux (thelio) — the PLATFORMS-append mechanism defeated.

Full `make check` green on this branch, including its own record→run→verify lockfile pass ("9 files, byte-identical").

Note: the ruby lockfile carries an extra `arm64-darwin-23` platform entry from the generating machine's bundler alongside generic `arm64-darwin`; harmless under frozen installs and byte-stable on both OSes.

---
## Summary by cubic
Improves build hygiene: `make clean` no longer deletes tracked files, Ruby JSON reads are pinned to UTF-8, and runner lockfiles are tracked with frozen installs for reproducible CI and fast drift detection.

- **Bug Fixes**
  - `go/Makefile` stops removing tracked `pkg/generated/client.gen.go`; added explicit `clean-generated`/`go-clean-generated` for opt-in rebuilds.
  - All Ruby `File.read` calls that parse JSON now use `encoding: "UTF-8"` (library, runners, tests); CI runs Ruby 3.3 with `LC_ALL=C` to prevent regressions.

- **Build/CI**
  - Track `conformance/runner/ruby/Gemfile.lock` and `conformance/runner/python/uv.lock`; installs run frozen via `BUNDLE_FROZEN=true` and `uv sync --locked` in Makefiles and CI.
  - Updated release checks and guidance to cover `npm ci`, `BUNDLE_FROZEN=true bundle install`, `uv sync --locked`, and `go` tidy; removed ignores for the lockfiles.

<sup>Written for commit 1c73cf68bbd79f9bd9634875862f12a6dd6425c8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/basecamp/basecamp-sdk/pull/712?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -531,6 +531,10 @@ jobs:
       - name: Run Ruby conformance tests
         if: matrix.ruby == '3.3'
         working-directory: conformance/runner/ruby
+        # Frozen: the runner Gemfile.lock is tracked (#670), so the install
+        # must fail fast rather than rewrite it.
+        env:
+          BUNDLE_FROZEN: 'true'
         run: |
           bundle install
           ruby runner.rb
@@ -626,8 +630,10 @@ jobs:
       - name: Run Python conformance tests
         if: matrix.python == '3.13'
         working-directory: conformance/runner/python
+        # --locked: the runner uv.lock is tracked (#670), so a missing or
+        # out-of-date lockfile must fail the install, not regenerate it.
         run: |
-          uv sync --python ${{ matrix.python }}
+          uv sync --locked --python ${{ matrix.python }}
           uv run --python ${{ matrix.python }} python runner.py
 
       - name: Run Python conformance runner unit tests
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -517,6 +517,15 @@ jobs:
         run: ../scripts/check-ruby-service-drift.sh
 
       - name: Test
+        if: matrix.ruby != '3.3'
+        run: bundle exec rake test
+
+      # File reads are pinned to encoding: "UTF-8"; this leg proves it stays
+      # that way by running the suite under the C locale (#702).
+      - name: Test (LC_ALL=C)
+        if: matrix.ruby == '3.3'
+        env:
+          LC_ALL: C
         run: bundle exec rake test
 
       - name: Run Ruby conformance tests
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,6 @@ spec/build/
 # Conformance test runner artifacts
 conformance/runner/go/conformance-runner
 conformance/runner/go/go
-conformance/runner/ruby/Gemfile.lock
 conformance/runner/typescript/node_modules
 # Vitest writes its results cache under whichever node_modules is closest to the
 # cwd, which is not always a package directory — ignore the tree everywhere so a
@@ -57,7 +56,6 @@ scripts/*.pyc
 conformance/runner/python/.venv/
 conformance/runner/python/**/__pycache__/
 conformance/runner/python/**/*.pyc
-conformance/runner/python/uv.lock
 
 # TypeScript SDK build artifacts
 typescript/coverage/
--- a/Makefile
+++ b/Makefile
@@ -307,7 +307,7 @@ doc-constants-check:
 # Go SDK targets (delegates to go/Makefile)
 #------------------------------------------------------------------------------
 
-.PHONY: go-test go-lint go-check go-clean go-check-drift go-check-wrapper-drift go-check-generated-drift check-grouped-client-coverage test-check-grouped-client-coverage
+.PHONY: go-test go-lint go-check go-clean go-clean-generated go-check-drift go-check-wrapper-drift go-check-generated-drift check-grouped-client-coverage test-check-grouped-client-coverage
 
 go-test:
 	@$(MAKE) -C go test
@@ -320,6 +320,11 @@ go-check:
 
 go-clean:
 	@$(MAKE) -C go clean
+
+# Remove the TRACKED generated Go client so `make -C go generate` rebuilds it.
+# Deliberately not part of `clean`: clean must never delete tracked files (#668).
+go-clean-generated:
+	@$(MAKE) -C go clean-generated
 
 # Check for drift between generated client and service layer
 go-check-drift:
@@ -1373,6 +1378,7 @@ help:
 	@echo "  test-check-grouped-client-coverage  Self-test that gate with adversarial inventories"
 	@echo "  go-check-generated-drift Check generated client.gen.go is current (regenerate + diff)"
 	@echo "  go-clean         Remove Go build artifacts"
+	@echo "  go-clean-generated  Remove the tracked generated Go client (explicit opt-in)"
 	@echo ""
 	@echo "TypeScript SDK:"
 	@echo "  ts-generate           Generate types and metadata from OpenAPI"
--- a/Makefile
+++ b/Makefile
@@ -238,12 +238,11 @@ endif
 		{ echo "ERROR: python/uv.lock is stale. Run 'make bump VERSION=$(VERSION)' first."; exit 1; }
 	@test "$$(jq -r '.packages["../../../typescript"].version' conformance/runner/typescript/package-lock.json)" = "$(VERSION)" || \
 		{ echo "ERROR: conformance/runner/typescript/package-lock.json records a stale SDK version. Run 'make bump VERSION=$(VERSION)' first."; exit 1; }
-	@# The conformance Ruby/Python runner lockfiles are gitignored and absent on a
-	@# fresh clone — check them only where they exist (a machine that has run
-	@# conformance), which is exactly where a stale one breaks `make check` (#671).
-	@test ! -f conformance/runner/ruby/Gemfile.lock || grep -qF 'basecamp-sdk ($(VERSION))' conformance/runner/ruby/Gemfile.lock || \
+	@# The conformance Ruby/Python runner lockfiles are tracked (#670), so they
+	@# are always present; a stale one breaks the frozen conformance installs (#671).
+	@grep -qF 'basecamp-sdk ($(VERSION))' conformance/runner/ruby/Gemfile.lock || \
 		{ echo "ERROR: conformance/runner/ruby/Gemfile.lock records a stale SDK version. Run 'make bump VERSION=$(VERSION)' first."; exit 1; }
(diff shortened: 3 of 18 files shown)


jeremy:
@codex review

Issue #708, opened 2026-08-12
Flaky on CI: proxy CONNECT drip deadline test lost its wall-clock race (5.0s observed vs < 2.0 asserted)


jeremy:
`OAuthTransportTest#test_proxy_connect_drip_is_bounded_by_the_deadline` (`ruby/test/basecamp/oauth_transport_test.rb:927`) failed on the Ruby 3.2 job of a docs-only PR (#707, run 31559343756): `Expected 5.038225341 to be < 2.0`. The test asserts the ~0.5s CONNECT-parse deadline cuts a drip-fed response by measuring wall clock with a 2.0s ceiling — a slow shared runner blew through it.

Same class as #655: a real-time assertion racing scheduler latency on CI. Nothing in the PR touches code (spec/api-gaps prose only). Tracking here so the next occurrence has a home — the likely fix direction is the same as #655's discussion (widen the ceiling, or assert the mechanism rather than the elapsed time). Rerun result will be noted below.

jeremy:
Rerun of run 31559343756 passed with no changes — confirming the failure was runner timing, not the PR.

PR #734, opened 2026-08-13
Stop reporting two environmental faults as red builds

### LINKED ISSUE #720: Flaky under parallel make: OAuthTransportTest#test_https_proxy_refused_without_p_use_ssl_support does live DNS inside a 1s bound
## What happens

`ruby/test/basecamp/oauth_transport_test.rb:839` —
`OAuthTransportTest#test_https_proxy_refused_without_p_use_ssl_support` — was
observed failing during a parallel `make` run while the #669/#674 build-hygiene
wave was being proven (2026-08-12, macOS, several concurrent agent workspaces on
one machine). It passes on a quiet machine and it passed on every serial run
since. Filing it rather than fixing it, because it is out of scope for the PR
that hit it and it should not be lost.

I have not reproduced it deterministically. What follows is a mechanism read of
the source, which is the actionable part.

## Why it is timing-dependent

The test asserts the fetcher refuses an `https://` proxy when the bundled
net-http lacks `Net::HTTP.new`'s eighth `p_use_ssl` parameter:

```ruby
error = assert_raises(Basecamp::Oauth::OauthError) do
  Basecamp::Oauth::Fetcher.stream_http(:get, "https://old-net-http.test/token", timeout: 1)
end
assert_equal "validation", error.type
```

That `OauthError` is raised at `ruby/lib/basecamp/oauth/fetcher.rb:371`. Reaching
it requires getting past line 363 first:

```ruby
proxy_uri = Timeout.timeout(timeout, Net::OpenTimeout) { uri.find_proxy }
```

`URI::Generic#find_proxy` resolves the target hostname — `IPSocket.getaddress` —
to evaluate its loopback rule. The fetcher's own comment above that line says so
and is deliberate about it ("a blocking DNS call that must sit inside the
advertised bound like every other network step"). So the test performs a **live
DNS lookup for `old-net-http.test` inside a 1-second bound**, before the
assertion's subject is ever evaluated.

`.test` is RFC 6761 reserved, so this is normally a fast NXDOMAIN — but it is
still a real query to whatever resolver the machine has. Blow the second and the
`Timeout` fires instead, surfacing as `Faraday::TimeoutError` rather than
`Basecamp::Oauth::OauthError`, and `assert_raises` fails. Under `make -j` the
same machine is running npm, uv, bundler, Gradle and Go concurrently, which is
exactly when a resolver round-trip stops being free.

That predicts a specific failure signature. Worth confirming against the real
failure output before acting:

```
Expected Basecamp::Oauth::OauthError, got #<Faraday::TimeoutError>
```

If the observed failure was a different error or a different assertion, this
analysis is wrong and the finding still stands on the observation.

## Notes for whoever picks it up

- The neighbouring test at `oauth_transport_test.rb:800` already stubs
  `IPSocket.getaddress` (there, to *simulate* a stall). This test does not, so it
  is the only one in the file whose pass/fail depends on the machine's resolver.
- The `timeout: 1` is doing double duty: the test's subject is a validation
  refusal that needs no timeout at all, but the value also bounds a DNS call the
  test does not care about.
- Directions, not a prescription — which is right depends on what the test is
  meant to pin: stub the resolution the way its neighbour does; raise the bound;
  or reorder so the capability check precedes proxy resolution (that last one
  changes shipped behaviour and would need its own argument).
- `ruby/test` is not parallelized (no `parallelize_me!`), so this is not
  intra-suite interference. The concurrency is `make -j` across toolchains,
  competing for the machine — not for Ruby's process.

## Scope

Split off from the PR closing #669 and #674. That PR closes #674 against its
**Gradle half only** (serializing the Gradle-backed `check-targets`); this is the
other thing `make -j` surfaced, and it is unrelated to Gradle.

### LINKED ISSUE #729: Both Gradle wrappers set retries=0, so a transient services.gradle.org 503 fails the job
Both Gradle wrappers set `retries=0`, so any transient failure fetching the Gradle distribution fails the job outright.

`kotlin/gradle/wrapper/gradle-wrapper.properties` and `spec/smithy-bare-arrays/gradle/wrapper/gradle-wrapper.properties` are identical on this point:

```properties
distributionUrl=https\://services.gradle.org/distributions/gradle-9.7.0-bin.zip
networkTimeout=10000
retries=0
retryBackOffMs=500
```

The retry machinery is already wired — `retryBackOffMs` is set and waiting for a non-zero `retries`. It's a one-line change per file.

## Observed rate

Four failures in a single afternoon's CI across three unrelated PRs, all the same error:

```
Attempt 1/1 failed. Reason: Server returned HTTP response code: 503 for URL: https://services.gradle.org/distributions/gradle-9.7.0-bin.zip
##[error]Exception in thread "main" java.io.IOException: Server returned HTTP response code: 503 ...
```

- #724 — `Kotlin Tests`, and separately `Gradle Dependency Check (Kotlin SDK)` twice
- #725 — `smithy / Verify Smithy spec`

`Attempt 1/1` is the wrapper reporting its own configured ceiling. Each one cost a manual `gh run rerun --failed` and a full job re-queue for a fault that a 500ms backoff would almost certainly have absorbed.

## Why this is worth fixing rather than tolerating

The failure is indistinguishable at a glance from a real Kotlin or Smithy break — same red X, same job name. Every occurrence costs someone a log read to classify it, and the classification is always the same. It also trains the reflex of re-running red Gradle jobs without reading them, which is exactly how a genuine failure gets waved through.

Note this is a *download* retry, not a build retry: it re-fetches the distribution zip on a 503. It cannot mask a flaky test or a real compile error, so it buys reliability without hiding signal.

## Suggested change

Set `retries=3` in both files, keeping `retryBackOffMs=500`. Consider raising `networkTimeout` from 10s at the same time, though the observed failures are immediate 503s rather than timeouts, so that part is speculative and could be left alone.

Both files are generated by `gradle wrapper` — check whether the repo regenerates them, in which case the setting needs to survive regeneration rather than be hand-patched.

jeremy:
Two CI-reliability fixes. They share a harm, not a mechanism: today CI reports
an environmental fault with the same red X as a real failure, which trains the
reflex of re-running red jobs without reading them.

Closes #729
Closes #720

---

## #729 — both Gradle wrappers set `retries=0`

`kotlin/` and `spec/smithy-bare-arrays/` both carried the Wrapper task's
default `retries=0`, so the wrapper's download loop got exactly one attempt:
`Attempt 1/1 failed` and the job was red. Set to 3 in both — four attempts,
500/1000/2000ms backoff.

**The files are generator output, so setting the line was not enough.** There
is no `gradle wrapper` invocation anywhere in the repo today, but the Wrapper
task exposes `getRetries()`, so both root builds now configure `tasks.wrapper`
and the setting survives regeneration.

`networkTimeout` stays at 10s, deliberately. Every observed fault was
immediate — a 503 or an EOF, never a `SocketTimeoutException` — so raising it
treats a case we have not seen, while multiplying what a genuinely hung
endpoint costs (4 attempts × 10s + 3.5s backoff instead of one 10s wait).

This retries the **distribution download**, not the build. It cannot mask a
flaky test or a real compile error.

### What is proven

Read out of the shipped `gradle-wrapper.jar` bytecode:

- `WrapperExecutor` reads `retries`/`retryBackOffMs` from the properties file;
  `WrapperConfiguration`'s defaults are `0` and `500`.
- `Install.forceFetch` loops `retries + 1` times, doubling the backoff, and
  logs `Attempt %d/%d failed. Reason: %s` — the exact string in #729's logs.
- The whole download sits in one `catch (IOException)`. The 503 surfaces from
  `Download.download`'s `getResponseCode()` *inside* that region, which is why
  #729's log shows the loop's own message before the rethrow.
- `validateDistributionUrl=true` adds no separate un-retried network call
  (checked empirically, `gradle-retry-validate-on.log`).

A/B against a blackholed distribution URL (`gradle-retry-ab.log`):

```
retries=0 → Fetching distribution.
            Attempt 1/1 failed. Reason: Connection refused
retries=3 → Fetching distribution (retrying 3 times, with an initial back off of 500 ms).
            Attempt 1/4 … Attempt 4/4 failed. Reason: Connection refused
```

Regeneration survival, with a negative control:

| log | result |
|---|---|
| `gradle-regen-smithy.log` | `./gradlew wrapper` reproduces the committed file byte-identically, `REAL_EXIT=0` |
| `gradle-regen-kotlin.log` | same, and `gradlew` + `gradle-wrapper.jar` byte-identical too |
| `gradle-regen-smithy-NEGATIVE.log` | with the `tasks.wrapper` block stripped, regeneration silently rewrites `retries=3` → `retries=0` |

### What is NOT proven

**That this absorbs a real `services.gradle.org` 503.** A 503 cannot be
summoned on demand, and I did not fabricate one. The chain above shows the
loop's ceiling changed and that a 503 lands inside the retried region; what
remains unverified is whether a real outage clears inside 500/1000/2000ms of
backoff. If services.gradle.org is down for minutes, four attempts still fail —
just more slowly.

Dependabot's `gradle-wrapper` bumps have rewritten only `distributionUrl` in
every historical instance (three checked), so they preserve the setting; that
is evidence, not a guarantee about a tool we do not control.

---

## #720 — `OAuthTransportTest` timing/network fragility

The issue records two failures and reads them as two mechanisms. **They are
one, and it is not timing.**

`stream_http` resolves the proxy with `URI#find_proxy` *inside* the advertised
deadline (`fetcher.rb:363`), and `find_proxy` calls
`IPSocket.getaddress(hostname)` to evaluate its loopback rule. Every proxy test
in the file targets a `.test` name — which RFC 6761 reserves as never
resolvable — so that is a real round trip to a resolver that has to answer
NXDOMAIN, sitting inside a 0.5–1s budget.

The sweep found **four** such tests, not two:

| test | budget |
|---|---|
| `test_proxy_credentials_are_percent_decoded` | 1s |
| `test_https_proxy_refused_without_p_use_ssl_support` | 1s |
| `test_https_proxy_scheme_gets_tls_on_the_proxy_connection` | 1s |
| `test_proxy_connect_drip_is_bounded_by_the_deadline` | 0.5s |

Instrumenting `IPSocket.getaddress` on unmodified `main` shows all four making
real lookups (`ruby-dns-instrument-BEFORE.log`, `DNSPROBE_TOTAL_CALLS=4`).

### Red proof

Slow that one resolver to 1.2s and **both** recorded failures reproduce
verbatim, from a single injected cause (`ruby-RED-slow-resolver-BEFORE.log`,
`REAL_EXIT=1`):

```
1) Failure:
OAuthTransportTest#test_proxy_connect_drip_is_bounded_by_the_deadline [:970]:
the proxy never saw the client cut the CONNECT drip — the whole-phase connect bound did not fire

2) Failure:
OAuthTransportTest#test_https_proxy_refused_without_p_use_ssl_support [:853]:
[Basecamp::Oauth::OauthError] exception expected, not
Class: <Faraday::TimeoutError>
```

The first is character-for-character the failure recorded in #720's comment.

**This also explains why #708 did not hold.** #708 correctly moved the drip
test off wall-clock assertions to a mechanism assertion, and argued its 15s
wait bound was far outside any plausible scheduling delay. That argument was
right; the bound was never the problem. With a slow resolver the request dies
in `find_proxy`, the proxy never receives a CONNECT at all, and an **empty**
queue is indistinguishable from a late one no matter how long you wait. No size
of wait bound could have covered it.

### The fix widens nothing

`with_proxy_env` scrubs the proxy environment and stubs `getaddress` to a fixed
non-loopback address — the only thing `find_proxy` asks of it — so the resolver
leaves the deadline entirely. Every bound #708 set stays exactly as it was,
doing the hang-guard job it was meant for. Five open-coded ENV save/restore
blocks collapse into the helper.

`test_proxy_resolution_dns_is_inside_the_deadline` is *about* resolution, so it
opts out with `resolves_to: nil` and keeps its own stub.

Green under the same 1.2s resolver (`ruby-GREEN-slow-resolver-AFTER.log`,
`REAL_EXIT=0`): the resolver is never called at all, and the set runs in 1.34s
instead of 16.5s.

### The tests still catch their bugs

Deflaking must not defang. Each changed test was run against a mutated
`fetcher.rb` (`ruby-MUTATION-matrix.log`); all five die, each at its guarding
assertion, and the file was restored by copy and verified with `diff -q` after
every run:

| mutation | test | `REAL_EXIT` | failed at |
|---|---|---|---|
| drop the `https://` proxy refusal | `..._refused_without_p_use_ssl_support` | 1 | `assert_raises(OauthError)` — got `ConnectionFailed` |
| drop the whole-phase connect bound | `..._connect_drip_is_bounded_by_the_deadline` | 1 | `assert cut.pop` — the mechanism assertion |
| drop the credential percent-decode | `..._credentials_are_percent_decoded` | 1 | `Proxy-Authorization` (`p%40s+s` vs `p@s+s`) |
| drop `p_use_ssl` passthrough | `..._gets_tls_on_the_proxy_connection` | 1 | ClientHello byte (`Expected: 22`) |
| drop the `find_proxy` bound | `..._dns_is_inside_the_deadline` | 1 | `assert_raises` — got `ReadDeadlineExceeded` |

That last one contradicted a comment claiming the error class does not
discriminate there. The comment is corrected rather than left standing.

### What is NOT proven, and what I did not do

**That the recorded CI failures were caused by resolver latency.** The injected
cause reproduces both messages exactly, and it is by far the most economical
explanation, but no DNS timing was captured on those runs. If the real cause
was something else, this fix does not address it. What *is* proven is that
these tests no longer depend on the resolver at all — one whole class of
environmental input is gone from them.

**Repetition is weak evidence, reported as such.** 12/12 green under 2× CPU
saturation (`ruby-repeat-under-parallel-load.log`). That is not a proof and is
not offered as one; the mutation matrix and the red/green pair are the
argument.

**Deliberately not addressed** — the sub-second `assert_operator seconds, :<,
TIMEOUT` bounds elsewhere in the file (`test_device_auth_non_2xx_...`,
`test_discovery_non_2xx_...`, and siblings). These have the same *shape* but a
different *property*: the margin between correct (~10ms) and broken (the
deadline itself) **is** the deadline, so widening the bound deletes the
discrimination rather than protecting it, and no mechanism assertion is
available to replace it. They stay coupled to real timing. None has been
observed failing. Recording the reasoning here so it is not re-litigated.

Also left alone: `test_proxy_credentials_are_percent_decoded` reads `captured`
across threads without explicit synchronization. It is ordered in practice —
the proxy appends every header line before the `close` that makes the client
raise — so there is a happens-before edge; flagging it rather than churning it.

---
## Summary by cubic
Stops two environmental faults from showing as red builds. Gradle wrapper now retries transient distribution downloads; OAuth proxy tests no longer depend on live DNS inside their deadlines.

**Gradle wrapper retries**
- Change: `retries=0` → `retries=3` with `retryBackOffMs=500` (four attempts). `networkTimeout` stays 10s.
- Persist via `tasks.wrapper` in both `kotlin/` and `spec/smithy-bare-arrays/` so `./gradlew wrapper` re-emits the setting.
- Scope: only the distribution download; does not mask real build or test failures.
- Review: confirm identical properties edits and identical `tasks.wrapper` blocks in both builds.

**OAuth proxy tests**
- Add `with_proxy_env` helper that scrubs proxy env and stubs `IPSocket.getaddress` to a fixed non-loopback address, removing DNS from proxy deadlines.
- Convert four proxy tests to use the helper; keep `test_proxy_resolution_dns_is_inside_the_deadline` opting out with its own resolver stub.
- Keeps existing time bounds as hang guards; eliminates flakes from resolver latency.
- Review: verify `IPSocket.getaddress` stub/restore and env handling. No runtime code changes or migrations.

<sup>Written for commit f57a753900af0390dc3f5868ed81eb7b25413cd0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/basecamp/basecamp-sdk/pull/734?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jeremy:
@codex review

jeremy:
## Rebased onto `9e8859b7c` (#724) — re-verification

New head: `f57a753900af0390dc3f5868ed81eb7b25413cd0`

### How the build files resolved

Both conflicts resolved **additively**; #724's blocks are kept verbatim and
untouched, with the wrapper block appended after.

- `kotlin/build.gradle.kts` — the only textual conflict. #724's `subprojects { tasks.withType<JavaCompile> { options.encoding = "UTF-8" } }` occupies the same append point. Its 21-line prefix is now **byte-identical to `9e8859b7c`**; `git diff 9e8859b7c` on this file is a pure addition of my block.
- `spec/smithy-bare-arrays/build.gradle.kts` — auto-merged (#724 inserted after the `java {}` block, mine after `tasks.test {}`). `git diff 9e8859b7c` is likewise a pure addition.

Nothing was reordered or restructured, so a reviewer diffing against `9e8859b7c` sees only the added block in each file.

### Regeneration proof, re-run against the rebased tree

The interaction #724 creates is worth naming: `./gradlew wrapper` now runs against a build file that carries *both* configurations. I widened the proof to answer it — the source is `git archive HEAD` (**tracked files only**, no stray build artifacts) and the comparison is a **whole-tree** `diff -r`, so any file the Wrapper task rewrote would surface, not just the properties file.

| proof | `REAL_EXIT` | tree diff |
|---|---|---|
| `REBASED-regen-kotlin-rebased.log` | 0 | **0** — entire tracked tree byte-identical |
| `REBASED-regen-smithy-rebased.log` | 0 | **0** — entire tracked tree byte-identical |
| `REBASED-regen-smithy-rebased-NEGATIVE.log` | 0 | 1 — `-retries=3` / `+retries=0` |

**The encoding pin survives regeneration**, and by the strongest available argument: the Wrapper task rewrote *nothing* in the tracked tree, `build.gradle.kts` included. It only ever writes `gradlew`, `gradlew.bat`, the jar and the properties file, and all four came back identical.

The negative control still discriminates, and it discriminates on **my** block specifically: stripping only `tasks.wrapper` — with #724's encoding pin verified still present (`encoding_pin_still_there=1`) — makes regeneration rewrite `retries=3` back to `0`.

### One pre-existing thing the whole-tree diff surfaced

`./gradlew wrapper` drops an **untracked `gradlew.bat`** in both modules. Neither module tracks or gitignores it. This predates both PRs and is unrelated to either change — it is simply what the Wrapper task always emits — but since this PR makes wrapper regeneration a documented path, it is worth knowing that regenerating leaves an untracked file behind. Whether to commit `gradlew.bat` is a Windows-support decision, not this PR's, so I have left it alone.

### Ruby proofs, re-run (unaffected, but confirmed rather than assumed)

| proof | `REAL_EXIT` |
|---|---|
| DNS instrumentation on the fixed tests | 0 — **`DNSPROBE_TOTAL_CALLS=0`** |
| green under the 1.2s slow resolver | 0 — resolver never consulted |
| whole file (44 runs) | 0 |
| rubocop | 0 |
| mutation matrix, all 5 | **1** each, all `RESTORED=OK` |

`DNSPROBE_TOTAL_CALLS=0` is the same instrumentation that counted **4** real lookups on unmodified `main`.

### Full `make` on the rebased tree

`REAL_EXIT=0`, `==> All checks passed`, working tree clean afterwards. This matters more than a normal re-run: #724 reworked the Makefile's Gradle-backed check targets with new order-only serialization edges, so this is the first time those edges and the wrapper config have run together.

jeremy:
@codex review

PR #748, opened 2026-08-17
Wire Python's last two generated services, and test the inventory

### LINKED ISSUE #732: Python's Client exposes 51 of 53 generated services — gauges and my_notifications are unreachable
Surfaced while writing the Python gauge tests for #719.

`python/src/basecamp/generated/services/` holds 53 service modules. `Client` and `AsyncClient` wire 51 of them. Two have no property at all:

```
$ ls python/src/basecamp/generated/services/*.py | xargs -n1 basename | sed 's/\.py$//' \
    | grep -v '^_' | sort > /tmp/gen.txt
$ rg -n '_service\("' python/src/basecamp/client.py | sed 's/.*_service("\([a-z_]*\)".*/\1/' \
    | sort > /tmp/wired.txt
$ comm -23 /tmp/gen.txt /tmp/wired.txt
gauges
my_notifications
webhooks_service      # false positive: wired as `webhooks`, module name differs
```

Both are missing from `async_client.py` too. `git log -S'gauges' -- python/src/basecamp/client.py` is empty, so `gauges` was never wired — `GaugesService` has shipped unreachable since `7216b767a` (#221).

Every other SDK exposes both:

| SDK | gauges accessor |
|---|---|
| TypeScript | `typescript/src/client.ts:521` |
| Ruby | `ruby/lib/basecamp/client.rb:578-581` |
| Kotlin | `generated/ServiceAccessors.kt:111` |
| Swift | `Generated/AccountClient+Services.swift:29` |
| Go | `Gauges()` on the account client |
| **Python** | **none** |

AGENTS.md's completeness bar names client wiring as a required step for every operation, so this is a bar violation on seven operations (`ListGauges`, `ListGaugeNeedles`, `GetGaugeNeedle`, `CreateGaugeNeedle`, `UpdateGaugeNeedle`, `DestroyGaugeNeedle`, `ToggleGauge`) plus whatever `my_notifications` carries.

## Why it went unnoticed

`make py-check-drift` compares the generated service layer against the OpenAPI spec — it never asks whether `Client` reaches what was generated. Nothing else does either. There is an established workaround that quietly absorbs the gap: `python/tests/services/test_my_notifications.py` instantiates `MyNotificationsService` directly rather than going through the client, so the unreachability never shows up as a failing test.

The gauge tests added in #719 follow that same precedent, which is why they pass today. That is the wrong shape long-term — a test that constructs the service by hand proves nothing about whether a user can get to it.

## Suggested fix

Add the two properties to `client.py` and `async_client.py`, then add a gate so the next one cannot slip: a test asserting that every non-underscore module under `generated/services/` has a corresponding accessor (with an explicit, reasoned allowlist for any deliberate omission). Without the gate this recurs — it is the third silently-unwired service by module count and nothing observed the first two.

Note the interaction with #600 (SPEC §5's account-scoped service roster is five short): whatever number that settles on should be the number this gate asserts, so the two do not drift apart.

jeremy:
`gauges` and `my_notifications` had no accessor on `AccountClient` or `AsyncAccountClient` — 51 of the 53 generated services were reachable. Both service modules ship, both are exported from `basecamp.generated.services`, and neither client ever named them, so `GaugesService` has been unreachable since #221.

Nothing observed the gap. `make py-check-drift` compares the generated service layer against the OpenAPI spec; it never asks whether the client reaches what was generated. The two tests that touched these services constructed them by hand over an account client, which passes whether or not a user could get there.

## The test is the larger half

`TestGroupedClientAccessorInventory` in `python/tests/test_client.py` (which had no accessor coverage at all) derives its roster from `basecamp/generated/services/` rather than a hand-copied literal — a newly generated service enters the test the moment it is generated and fails until someone wires it. A literal list would have to be edited by the same person who forgot the accessor, which is how this shipped.

One spelling rule is encoded as data: `webhooks_service.py` → `webhooks` (`generate_services.py`'s `service_filename` renames it to avoid clashing with the `basecamp.webhooks` package). Nothing else is normalized. The sync/async service classes are read off each module with `inspect` rather than re-derived from the filename, so no class-name spelling is restated either.

Four things are checked:

1. every generated module has a sync accessor
2. every generated module has an async accessor
3. the accessor resolves to *that module's own* service class, sync paired with sync and async with async — presence alone would pass if `gauges` returned the wrong service
4. the other direction: no accessor survives its generated service

Plus a guard that the derived roster is non-empty, so a path change can't make the rest vacuously true.

## Red proofs

Each case was shown to fail against un-fixed code. Every mutation was applied to a real input, verified on disk, and restored by `cp` + `diff -q`.

| Mutation | Result |
|---|---|
| both clients reverted to `origin/main` (51 accessors, verified on disk) | 10 failed — presence ×4, resolve ×4, no-orphans ×2 |
| `gauges` accessor points at `GoogleDocumentsService` | 1 failed: `test_accessor_resolves_to_that_module_s_service[gauges]` — presence still passed, which is the point |
| async `my_notifications` handed the *sync* class | 1 failed: `test_async_accessor_resolves_to_the_async_service[my_notifications]` |
| invented `gadgets` accessor with no generated service | 1 failed: `test_no_accessor_without_a_generated_service[AccountClient]`, naming `gadgets` |
| roster derivation pointed at an empty path | 3 failed, including the non-empty guard (`assert 0 > 40`) |

## Also

- The two hand-constructing tests now go through `account.gauges` / `account.my_notifications`, and their docstring caveats are gone.
- SPEC.md Appendix F's Python row: 51 → 53, noting it is now held by a test rather than by the hand-verification date.
- `python/README.md`'s paragraph telling users to construct unwired services themselves is deleted.

`LC_ALL=C make py-check` passes (1486 passed, 4 skipped; mypy, ruff, format clean), as do `doc-constants-check` and `check-readme-env-vars`.

Closes #732

---
## Summary by cubic
Wires Python’s last two generated services onto the account clients and adds an accessor-inventory test so all generated services stay reachable. Previously, `gauges` and `my_notifications` existed under `basecamp.generated.services` but had no accessors; now both `AccountClient` and `AsyncAccountClient` expose `account.gauges` and `account.my_notifications`. Closes #732.

**Review notes**
- Adds `gauges` and `my_notifications` properties to `python/src/basecamp/client.py` and `python/src/basecamp/async_client.py`, each returning the correct sync/async service class from `basecamp.generated.services`.
- Introduces `TestGroupedClientAccessorInventory` in `python/tests/test_client.py`, which derives the roster from `basecamp/generated/services/`, checks sync and async accessor presence, resolves to the module’s own service classes, and asserts no orphan accessors; includes the `webhooks_service` → `webhooks` mapping.
- Updates service tests to use `account.gauges` and `account.my_notifications` instead of hand-constructed services.
- Updates `SPEC.md` (Python row to 53, now guarded by the test) and simplifies `python/README.md` (removes the “construct unwired services” guidance).

**Migration**
- Replace direct construction with accessors:
  - `GaugesService(account)` → `account.gauges`
  - `MyNotificationsService(account)` → `account.my_notifications`

<sup>Written for commit 2e7d24ddac19f51815a8d1dfa05b45007cd52d49. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/basecamp/basecamp-sdk/pull/748?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -3647,8 +3647,11 @@ For ASCII text (all conformance test fixtures today), these are equivalent.
 Counts are of accessors actually wired onto the client, against §5's canonical
 roster. The Kotlin and Swift rows are marked, because §5's roster is derived from
 exactly those two files and a restatement of a gated value has to be gated too.
-The other four are hand-verified and dated below — no gate derives them, and
-building one means a sixth hand-copy of each generator's split tables.
+Of the other four, Python's is held by its own accessor-inventory test, which
+derives the roster from `python/src/basecamp/generated/services/` and fails when
+an accessor is missing. The remaining three are hand-verified and dated below —
+no gate derives them, and building one means a sixth hand-copy of each
+generator's split tables.
 
 | SDK | Account-scoped services |
 |-----|------------------------|
@@ -3657,11 +3660,13 @@ building one means a sixth hand-copy of each generator's split tables.
 | Ruby | 53 — full canonical set |
 | TypeScript | 53 — full canonical set, on the flat client alongside `authorization` (no `AccountClient` tier; see Client Topology above) |
 | Go | 51 accessors. Two services are folded rather than missing: `automation`'s sole operation is `LineupService.ListMarkers`, and `clientVisibility`'s is `RecordingsService.SetClientVisibility`. `timesheets` is spelled `Timesheet` (singular). Capability is 53/53; the surface is not. Hand-written service wrappers around the generated OpenAPI client — not fully generated. |
-| Python | 51 — `gauges` and `myNotifications` are a genuine wiring gap, not a fold: `generated/services/gauges.py` and `generated/services/my_notifications.py` both exist and neither `client.py` nor `async_client.py` references them (#732). Sync and async agree exactly. |
+| Python | 53 — full canonical set. `gauges` and `my_notifications` were a wiring gap rather than a fold, and were wired in #732; the same change added an accessor-inventory test (`python/tests/test_client.py`) deriving its roster from `generated/services/`, so the next unwired service fails rather than going unnoticed. Sync and async agree exactly. |
 
 Verified 2026-08-13 against the accessor declarations in each SDK's client
 (Go `AccountClient` methods, Ruby `Client#for_account` accessors, TS
-`defineService` calls, Python `_service` properties).
+`defineService` calls). Python was re-verified 2026-08-16 when #732 wired the
+last two, and no longer relies on that date: its `_service` properties are
+checked against the generated package on every `make py-check`.
 
 ### Event Feed Connector Scenario Lane (§23)
 
--- a/python/README.md
+++ b/python/README.md
@@ -455,7 +455,7 @@ token.is_expired(buffer_seconds=60)  # True if expiring within 60s
 
 ## Services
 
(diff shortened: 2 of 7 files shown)


jeremy:
@codex review

Issue #163, opened 2026-03-09
Search API spec may diverge from upstream BC3 docs


jeremy:
## Observation

The Smithy spec models the search endpoint (`SearchInput` in `spec/basecamp.smithy:6182`) with:
- `query` (required string, `@httpQuery("query")`)
- `sort` (optional `SearchSortField`, `@httpQuery("sort")`)
- `page` (optional integer, `@httpQuery("page")`)

The upstream BC3 API documentation uses:
- `q` (not `query`) as the search query parameter name
- `page` and `per_page` for pagination (documented)
- No documented `sort` parameter — relevance ordering is the default

## Mismatches

1. **Parameter name**: Smithy uses `query`, upstream docs use `q`
2. **Undocumented param**: `sort` is modeled in Smithy but not documented upstream
3. **Missing param**: `per_page` is documented upstream but not modeled in Smithy

## Impact

- The `query` vs `q` parameter name difference means SDK-generated requests may use the wrong parameter name (needs verification against actual API behavior — BC3 may accept both)
- `sort` may be an undocumented parameter that happens to work today but isn't part of the stable API contract

## Context

Discovered while fixing zero-value optional param serialization. The codegen fix (omitting zero-value `sort` and `page`) restores correct relevance-ranked results, but the underlying spec should be verified against the actual BC3 API behavior.

## Action

1. Verify actual BC3 search endpoint behavior (`q` vs `query`, whether `sort` is accepted)
2. Update `spec/basecamp.smithy` SearchInput to match upstream if diverged
3. Add `per_page` if supported
4. Regenerate all SDK clients after any spec correction

PR #851, opened 2026-09-09
Clear the open code-scanning alerts

jeremy:
## Scope

Every open CodeQL and Scorecard alert that is a code or workflow change. The remaining open alerts are governance settings (Scorecard #69 branch protection admin bypass, #83 approving reviews) and are not touched here; four Scorecard heuristics were dismissed on the alert with reasons (#222, #79 `npm install` in the release lockfile rewriter; #85 fuzzing; #84 best-practices badge).

| Alert | Rule | Verdict | Change |
|---|---|---|---|
| #232, #233 | CodeQL `py/uninitialized-local-variable` — `async_auth.py`, `auth.py` | false positive on a deliberate pattern; restructured so the analysis can prove it | one binding for both arms, narrow on it |
| #234, #235 | same — `oauth/device.py` | same | same |
| #236, #237 | same — `oauth/exchange.py` | same | same |
| #239 | CodeQL `actions/syntax-error` on `ruby/Gemfile.lock` | false positive: the actions analysis parsed a lockfile as YAML | `**/Gemfile.lock` in `paths-ignore` |
| #163 | Scorecard PinnedDependencies — `go install govulncheck@latest` | fix | pinned to `v1.8.0`; the vulnerability database is fetched live regardless |
| #70, #71 | Scorecard BinaryArtifacts — two `gradle-wrapper.jar` | jars match Gradle's published 9.7.1 SHA-256; setup-gradle already validates them, but Scorecard recognises only the dedicated action | `gradle/actions/wrapper-validation` job in `test.yml`, same SHA as the existing `setup-gradle` pin |

## The Python change

The six sites are SPEC §9 transport-failure handlers: the error is constructed inside `except` and raised after it so the httpx exception (which retains the request, whose form body carries `client_secret` / `refresh_token` / `device_code`) never lands in `__context__`. Before, the try arm bound `response` (or `status, body`) and the except arm bound `transport_error`, and the raise was guarded on `transport_error is not None`. That is correct, but a definite-assignment analysis cannot correlate the two names. Now both arms bind `outcome: Response | NetworkError` (or `tuple[int, bytes] | OAuthError`) and the continuing path is `if isinstance(outcome, Error): raise outcome`. Control flow, message text and no-chaining semantics are unchanged; mypy narrows the union without annotations.

Not changed: `device.py`'s token poll (`poll_error`) keeps its shape; CodeQL did not flag it and it has a `continue` arm that the union form would not simplify.

## Verification

- `make py-check` (1603 passed, mypy clean, ruff clean); full `make check` locally, see below.
- `actionlint` clean; `zizmor .` reports only the seven pre-existing `self-repository` lows on `uses: ./.github/workflows/smithy-verify.yml`, none from the new job.
- Scorecard runs on push to `main` and weekly; #70/#71 clear once `test.yml` has a green run on the latest `main` commit, #163 on the next scan. The CodeQL alerts close on the next `main` analysis.

---
## Summary by cubic
Clears the remaining open CodeQL and Scorecard alerts that are code or workflow changes, without changing runtime behavior.

**CodeQL**
- Restructured the six SPEC §9 transport-failure handlers so both try/except arms bind one name and the continuing path narrows on it; control flow, messages, and no-chaining semantics are unchanged.
- Added `**/Gemfile.lock` to `paths-ignore`; the actions analysis was parsing the lockfile as YAML.

**Scorecard**
- Pinned `govulncheck` to `v1.8.0`; the vulnerability database is still fetched live at scan time.
- Added a `gradle/actions/wrapper-validation` job to `test.yml`; the wrapper jars match Gradle's published checksum and `setup-gradle` already validates them, but Scorecard only recognizes the dedicated action.
- Remaining alerts are governance settings and dismissed heuristics, not touched here.

<sup>Written for commit 3afef347e8e33e6c4d3b9b3bd04c5ae889e2cbaa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/basecamp/basecamp-sdk/pull/851?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -30,6 +30,9 @@ paths-ignore:
   # out to be noisy (py/unused-import in every generated service), so codeql.yml's
   # SARIF filter now strips generated/ except services/_ — keeping both
   # hand-written files scanned.
+  # The actions analysis extracted this lockfile and reported a YAML syntax
+  # error (alert #239); no language analyses lockfiles.
+  - '**/Gemfile.lock'
   # Test infrastructure
   - conformance/
   - kotlin/conformance/
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -146,8 +146,10 @@ jobs:
           # Scan against the newest 1.26.x, not whatever patch the runner has cached.
           check-latest: true
 
+      # Pinned for Scorecard PinnedDependencies; the vulnerability database is
+      # fetched live at scan time regardless of the tool version.
       - name: Install govulncheck
-        run: go install golang.org/x/vuln/cmd/govulncheck@latest
+        run: go install golang.org/x/vuln/cmd/govulncheck@v1.8.0
 
       - name: Run govulncheck
         run: govulncheck ./...
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,6 +27,20 @@ jobs:
         with:
           advanced-security: false
 
+  # setup-gradle already validates the wrapper jars on the Kotlin jobs;
+  # Scorecard's BinaryArtifacts check recognises only this dedicated action.
+  gradle-wrapper:
+    name: Gradle wrapper validation
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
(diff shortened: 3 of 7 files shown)


jeremy:
Converged: CI green (including the new Gradle wrapper validation job and all seven CodeQL analyzers, which report zero alerts on the merge ref), Copilot approval-recommended, Codex completed with no findings, one poll of silence. Nothing outstanding from my side; merge is yours.

PR #877, opened 2026-09-12
Card table templates, and a template library that says which kind it means

thehale:
Basecamp's template library holds three kinds of template, but its only address dated from when to-do lists were the only thing in it. Every call under `template_library` silently meant "to-do lists", so a script that wanted a card table template had nothing to call.

Absorbs the API surface from https://github.com/basecamp/bc3/pull/13312 into all seven SDKs.

## What this adds

- **Card table templates** can be listed and created.
- **Reading to-do list templates** moves to `/template_library/todolists.json`, the address Basecamp now serves it from.
- **Copying a template** can name the destination *project* and let Basecamp resolve the container, so a caller needs to know nothing about docks or to-do sets.
- **Templatifying** lifts an existing to-do list or card table in a project into the library, asynchronously, with its own status resource.
- **Creating an empty to-do list template** gains a call, so both kinds can be made from scratch.

## Breaking changes

Two, both mechanical, both written up in `MIGRATING.md`.

**`GetTemplateLibrary` is deleted, not deprecated.** It becomes `GetTemplateLibraryTodolists`, and the method renames with it (`getLibrary` → `getLibraryTodolists`, `get_library` → `get_library_todolists`, `GetLibrary` → `GetLibraryTodolists`). The old address still answers with a redirect, but it is no longer modelled.

Deleting rather than aliasing is this repo's stated policy, SPEC.md:2701: *"A rename is breaking and ships without a deprecated alias."* It has been followed consistently — `UpdateDocument` is gone, `UpdateScheduleEntry` became `ReplaceScheduleEntry` with no alias, and `CreateForwardReply`, `GetRecording` and `GetEverythingBoosts` were all removed outright. The five pre-existing `@deprecated` sites in the spec are all *members that the server still serves and you should stop using*, never operations.

**`TemplateLibrary` becomes `TemplateLibraryTodolists`**, now that it has a sibling in `TemplateLibraryCardTables`. Members are unchanged; rename the type where you name it.

## Commits

Each builds and passes conformance on its own.

1. **Model the card table fields Basecamp already sends** — `position`, `parent` and `public_link_url` have always been returned and were never modelled, so every SDK dropped them on decode. Pre-existing drift, closed on its own rather than arriving unexplained inside a feature.
2. **Fold a long path parameter's documentation instead of emitting it as code** — a path parameter description longer than one line produced a Ruby file that would not parse, breaking `require` of the whole Ruby SDK. Body and query descriptions were already folded; path was the one branch that skipped it. Latent on `main`, tripped by a description written for this work.
3. **Let people script against card table templates**
4. **Let people templatify an existing to-do list or card table**
5. **Write down what upgrading to the template library changes**

## Before this merges

**https://github.com/basecamp/bc3/pull/13312 must ship first.** `make bc3-route-parity` fails on this branch by design: the SDK declares routes Basecamp does not document at the current provenance pin. That is the sanctioned holding pattern, registered in `spec/api-gaps/template-library-card-tables.md`, and it clears at the repin with no SDK change, because the routes enter `spec/bc3-routes.json` from Basecamp's own `doc/api/sections/`.

**A reviewer running `make -k check` will see five failures. All are expected.**

| Failure | Why |
|---|---|
| `bc3-route-parity` | Deliberate, as above. Clears at the repin. |
| `test-bc3-route-parity` | Cascade of the same thing: its positive control runs the real allowlist through the real checker. |
| `swift-check-drift` | Environmental. The script passes host-absolute paths, which do not resolve inside a containerised Swift toolchain. Generation was verified clean container-natively; the committed Swift is byte-identical to what the generator produces. |
| `rs-deny` | `cargo-deny` not installed locally. Fails identically on `main`. |
| `event-feed-fixtures-check` | Missing local `uv`. Fails identically on `main`. |

The last three are not caused by this branch and can be confirmed by running them on `main`.

**Swift conformance is macOS-gated** and skips on Linux, so the Swift runner changes are generated-and-consistent but never compiled. They want a look on a Mac before release.

## Verification

Conformance across all seven runners: 235 Go, 237 Rust, 237 Python, 218 Ruby, 236 Kotlin, 291 TypeScript. Plus 1745 TypeScript and 1608 Python unit tests, and the full Go and Rust suites. Grouped-client coverage over 267 operations, idempotency parity, deprecation parity, fixture coverage, doc constants, api-gap validation.

## Worth reading before reviewing

- **A cross-SDK naming divergence was caught and fixed.** `GetTemplatification` generated as `templatification(...)` in TypeScript, Kotlin, Rust and Swift while Python and Ruby kept the `get_` prefix, because the naming algorithm elides `Get` when it matches the resource type. All seven now agree on `getTemplatification`.

## One finding this surfaced, deliberately not fixed here

A committed fixture can outlive the contract it describes and nothing notices. Injecting a field the spec no longer has, into both a conformance fixture and a manifest-tracked shared fixture, leaves Go, Rust and `check-fixture-coverage` all green: decoders ignore unknown keys by design, and the fixture guard checks required-field presence and type but never rejects a key the schema lacks.

The same class of bug bit basecamp-cli for real, where a test asserted against a response Basecamp had stopped producing and CI passed on a broken command.

The distinction worth building on is that a **live** response carrying an unknown key is Basecamp moving ahead of the spec and should pass, while a **committed fixture** carrying one is stale and should fail. Same shape, opposite meaning. Left out of this PR on purpose: it is repo-wide test hygiene, and folding it into a card-table-templates change would bury it.


thehale:
🤖 All five findings were real and all five are fixed. Two commits, not yet pushed.

**Go could not express an explicit `false`.** Right, and worse than "silently omits": the three settings were plain `bool` passed through `omitzero`, so a caller had no way to turn one off, while the other six SDKs sent it. They are `*bool` now, matching the existing `VisibleToClients` / `Paused` / `NotifyOnAnswer` request fields.

The reason this slipped is that nothing exercised it, so a conformance case now pins that `false` reaches the wire. Reverting the field to plain `bool` makes that case fail, which is how it earns its place. `AddingPeopleConfirmed` has the same shape but predates this branch, so it is left alone rather than widening the change.

**`Templatification` was missing from `TYPE_ALIASES`.** Correct, and it is the second instance of that bug in this branch: `TemplateLibraryTodolists` and `TemplateLibraryCardTables` hit it earlier and were fixed, and this one was missed. Both calls return the model now instead of `JsonElement`, and the Kotlin conformance runner follows.

**Both operation counts were stale.** `267 − 91 = 176`, not 171, and `57 POSTs − 11 idempotent = 46`, not 43. Both were correct against the 262-operation surface and should have moved with the total. The `@operation-count` marker keeps the totals in step automatically; these breakdowns sit beside it and are hand-written, which is precisely why they drifted.

**The gap entry proposed an obsolete design.** Also correct. It still described a project field being added and the library read being retained and deprecated, both of which this branch does the opposite of. Rewritten to the six operations and the copy extension that actually ship, with an explicit note that no project field is added.

One thing worth recording rather than fixing: the count breakdowns are hand-maintained beside an auto-synced total, which is a shape that will drift again. Not changed here, since making them generated is repo-wide tooling and does not belong in a card-table-templates change.

jeremy (approved):

PR #906, opened 2026-09-16
Three Python errors crash with a bare TypeError when a caller passes retryable

jorgemanrubia:
Three of the Python SDK's error classes crash when you construct them with a `retryable` keyword:

```python
>>> RateLimitError(retryable=False)
TypeError: BasecampError.__init__() got multiple values for keyword argument 'retryable'
```

`NetworkError` and `LimitExceededError` do the same. `ApiError` does not — it takes the keyword and honours it.

Originally tracked in [37 Three Python SDK errors raise a bare TypeError on a retryable kwarg](https://app.basecamp.com/2914079/buckets/48699913/card_tables/cards/10307865976).

## Why this is worth fixing

An error class that raises an error *from outside the error taxonomy* defeats the point of having a taxonomy. A consumer catching `BasecampError` does not catch this. A consumer catching `Exception` catches it and cannot tell it from a bug in their own code. It is the same shape as [A null list body crashes the Python SDK instead of reading as an empty listing](https://github.com/basecamp/basecamp-sdk/pull/896), where a decode failure crossed a catch-class boundary and `except ValueError:` quietly stopped catching.

It is also not refused where a refusal would be useful. These constructors are typed `**kwargs: Any`, so the static contract says the call is fine:

```
$ mypy probe.py
probe.py:3: note: Revealed type is "basecamp.errors.RateLimitError"
Success: no issues found in 1 source file
```

mypy says yes; the runtime crashes. That is what makes this a defect rather than a house style.

## What should happen

Three answers were available, and the choice matters more than the diff: honour the caller's value, accept it and let the class invariant win, or refuse it as a `UsageError` inside the taxonomy. Silently raising `TypeError` was not one of them.

This takes the second — **the class invariant wins** — which is already what three siblings in the same hierarchy do. `CampfireDiscoveryIncompleteError`, `CampfireIndexLoadAbortedError` and the OAuth `DeviceFlowError` all overwrite the keyword rather than colliding with it, the first two as of [Summarize a recording from an event-feed pointer, in Python](https://github.com/basecamp/basecamp-sdk/pull/888), each with the reasoning written at the site.

Honouring the caller is the one to reject on the merits rather than on precedent. A 507 is never retryable: no amount of backoff frees storage, which is the entire reason `LimitExceededError` is not an `ApiError` and why the 507 arm is decided before the 5xx arms. 429 and a transport failure are retryable for the mirror-image reason. Refusing with a `UsageError` has no precedent in any of the six SDKs, and would make constructing an error a fallible operation that error-handling paths then have to handle.

`ApiError` honouring the caller is not an exception to that rule but the other half of it: it fixes no retryability at all (500 is retryable, 418 is not), so there is nothing for a caller's value to contradict. The rule: **where a class fixes its retryability, the class wins; where it does not, the caller's value is the value.** Both halves are now pinned per class by the sweep.

To be precise about what that rule does and does not claim: it governs how a constructor *treats the keyword*, not whether every class fixes the *right* retryability. It does not, and that is older than this change — `AuthError`, `ForbiddenError`, `NotFoundError` and `ValidationError` fix none in Python although SPEC §6 gives their codes `retryable: false`, and `PeopleConfirmationRequiredError` is `Omit<…, "retryable">` in TypeScript and takes no such kwarg in Ruby. Worth a separate look; not smuggled into this one.

One correction to the argument above, from the adversarial pass: `LimitExceededError` is not in this SDK's own retry catch tuple — `_http`/`_async_http` catch `(RateLimitError, NetworkError, ApiError)`, so a 507 never reaches `_is_retryable_error` here. The loop that would spin against a full disk is a *consumer's*. `.retryable` is a public answer consumers act on, which is what makes it worth defending; the source comment says that now rather than implying a hazard in our own retry path.

## The whole hierarchy, not the three

The card named three because that is where its author looked. Every `BasecampError` subclass was probed with every base keyword, passed *as a keyword* — the first version of this probe bound `message` positionally and so never tested it, which the adversarial pass caught. Corrected inventory, all 23 subclasses:

| keyword | collides on | verdict |
|---|---|---|
| `retryable` | the 3 fixed here | fixed |
| `message` | 5 composites that compose their message | left refused |
| `code` | 21 subclasses that fix one | left refused |
| `hint`, `http_status`, `retry_after`, `request_id` | none | — |

The two left alone are not the same defect, by the criterion this change turns on — whether a caller passing the keyword has an intent the crash defeats:

- **`code`** is the class's identity. `NotFoundError(code="rate_limit")` is a contradiction, not a request, and overwriting it would silently swallow a caller's real confusion.
- **`message`** on those five is *derived from the caller's own structured arguments* — `routing_key`, `bucket_id`, `reason`. The caller already controls it, through the arguments that compose it. There is no intent a refusal defeats.

`retryable` is neither: the base class declares it as an ordinary knob, `ApiError` honours it, and the value a caller asks for is not derivable from anything else they passed.

`WebhookVerificationError` advertises no keywords at all, so every stray one is "unexpected keyword argument" — refused by mypy as well as at runtime. Ordinary Python, not an error escaping the taxonomy.

The new tests include a sweep rather than a list, so a fourth class can reintroduce neither the crash nor the wrong answer: it walks the package, builds every `BasecampError` descendant, and requires that each answer the flag with the value it should — the invariant for the six classes that fix one, the caller's value for the rest. Every subclass must be classified, and the exact set of 23 names is pinned, so a class cannot arrive or leave unnoticed.

## MIGRATING.md

Not a break, and it gets a section anyway.

**Not a break:** every affected call raised `TypeError` **100% of the time**, on every input, so there is no working consumer code on this path, and nothing previously accepted changes value. The adversarial pass built both revisions and diffed 90 constructed errors plus `error_from_response` across 14 statuses, with pickle, `copy`, `__reduce__`, `functools.partial` aliasing and a forwarding subclass: every single difference is `RAISE → OK`. Strictly a widening. Label stays `bug`, not `breaking`.

**A section anyway**, because "did a working call change value" is not this repo's bar. `MIGRATING.md` already carries *"Ruby: three crash classes on server-supplied URLs became `ApiError` refusals"* — a pure crash-to-taxonomy widening whose own answer to "wrong behaviour if you ignore it" is **none**, documented regardless, with the line telling you what to rescue instead. And there is one genuinely silent edge the first draft of this PR talked past: code written against the crash now gets an object whose `.retryable` is the class's answer rather than the one it passed, with no signal the argument was dropped. That is the silent class the file exists for.

## Verification

Reverting any one of the three constructors alone fails `test_the_three_accept_the_flag_and_yield_their_invariant` and `test_every_subclass_answers_the_flag_with_the_value_it_should`. The other two new tests pass without the fix by design — they assert what did *not* change.

Regressing all six overwrite sites to `kwargs.setdefault` — "honour the caller", the semantics rejected above — also fails the sweep, by name and with the fix in the message. That is the regression an earlier version of the sweep let through green.

Full Python suite, mypy and ruff green locally and on CI: 2218 passed, 4 skipped, 94.42% coverage; 274 conformance cases passed.

Adversarial review: Opus.


jorgemanrubia:
🤖 @codex review

jorgemanrubia:
@codex review

jorgemanrubia:
🤖 **Reviewer record — round 1.** Adversarial pass with **Opus** on `e1b92a01`. No Codex pass: that integration is bound to the repository 'owner’s account and our requests are refused, so it is not a gate here and will not be asked for again. Same as [#888](https://github.com/basecamp/basecamp-sdk/pull/888), and recorded plainly rather than tidily. Copilot approved on `e1b92a01`.

Eight findings. **Five applied** in `6376f5e5`, **two answered**, **one corrected in both directions** — the reviewer was half right and the half it got wrong still exposed a false claim of mine.

---

### Applied

**1. The sweep swept the crash, not the semantics.** It asserted `isinstance(error.retryable, bool)`, which is true of every possible answer. Regressing all six overwrite sites to `kwargs.setdefault` — "honour the caller", the semantics this change rejects on the merits — left it green. Reproduced before fixing:

```
$ sed 's/kwargs\["retryable"\] = X/kwargs.setdefault("retryable", X)/'   # all 6 sites
$ pytest ...::test_no_subclass_that_advertises_kwargs_raises_on_the_flag -q
1 passed
```

This is the sharper version of the finding: [#888](https://github.com/basecamp/basecamp-sdk/pull/888)'s own commit message is a three-iteration post-mortem of a sweep *in this file* that swept nothing, and this was the same mistake one register down. The sweep now carries a per-class expected value — the six classes that fix retryability must answer with the invariant whichever way the flag is passed, every other class must answer with the caller's value, and every subclass must be in one bucket or the other. The same regression now fails by name:

```
E  CampfireDiscoveryIncompleteError fixes retryable=False, but a caller moved it:
   {True: True, False: False}. Overwrite the keyword -- never setdefault, which
   honours the caller instead.
```

That incidentally pins `ApiError`'s half of the rule, and `DeviceFlowError`, which no test pinned a value for before.

**2. Dead floor.** `>= 22` was implied by `>= 23` plus the one-element exempt set and could never fire on its own, and a floor only sees a removal until the next class arrives. Both replaced with the exact set of 23 class names, so addition and removal each force a decision.

**3. `__subclasses__()` saw more than this package.** It returns every subclass that has been *imported*, so an error defined at module scope in any other test file would join the walk — green under `pytest tests/test_errors.py`, red under the whole suite. Filtered to `basecamp.*`. Latent today; not tomorrow.

**4. A baseline failure was not actionable.** A new class with required arguments raised a bare `TypeError: missing 1 required positional argument` at the construction line while the docstring promised it would name the fix. It now fails with the class name and a pointer to the `required` map.

**5. `MIGRATING.md` was owed.** I argued none was needed because every affected call raised 100% of the time, so no working code reaches the path and no value moved. That is true — the reviewer's own differential (90 constructed errors plus `error_from_response` across 14 statuses, both revisions, with pickle, `copy`, `__reduce__`, `partial` aliasing and a forwarding subclass) found *every* difference to be `RAISE → OK`. But it is not this repo's bar: `MIGRATING.md` already carries "Ruby: three crash classes on server-supplied URLs became `ApiError` refusals", a pure crash-to-taxonomy widening whose own answer to "wrong behaviour if you ignore it" is **none**, documented anyway. And there is one genuinely silent edge I talked past — code written against the crash now gets an object whose `.retryable` is the class's answer rather than the one it passed, with no signal the argument was dropped. Section added. Still `bug`, not `breaking`.

---

### Corrected in both directions

**6. "Every base keyword was probed" was false — and the reviewer's count was wrong too.** It reported `message=` colliding on **11** classes. Six of those are an artifact of its probe, which bound `message` positionally *and* passed it as a keyword: `UsageError("m", message="x")` is a plain duplicate-argument error, and `UsageError(message="x")` works fine. The real number is **5** — the composites that compose their message from structured arguments.

But the part it got right is the part that mattered: my probe bound `message` positionally and so never tested it as a keyword, which means my claim to have "probed every base keyword" and found "two other shapes" was false as written. Corrected inventory, all 23 subclasses, every keyword passed *as* a keyword:

| keyword | collides on |
|---|---|
| `retryable` | the 3 fixed here |
| `message` | 5 composites |
| `code` | 21 |
| `hint`, `http_status`, `retry_after`, `request_id` | none |

`message` stays refused, and the reason is the same criterion that separates `retryable` from `code`: on those five it is *derived from the caller's own structured arguments* — `routing_key`, `bucket_id`, `reason` — so the caller already controls it and a refusal defeats no intent.

**7. The 507 comment overclaimed.** `LimitExceededError` is not in this SDK's own retry catch tuple — `_http`/`_async_http` catch `(RateLimitError, NetworkError, ApiError)`, so a 507 never reaches `_is_retryable_error` here. The loop that would spin against a full disk is a **consumer's**. `.retryable` is a public answer consumers act on, which is what makes it worth defending; the comment says that now instead of implying a hazard in our own retry path.

---

### Answered, not applied

**8. "Four of five sibling SDKs close the signature; you never considered that option."** The parity table is correct and useful — Ruby, TypeScript, Go and Swift all refuse the keyword *statically* by enumerating their constructor parameters; only Rust allows `.retryable(bool)` on every error including a limit-exceeded one. Closing the signature here would also be a strict widening-of-nothing, and would kill the `code=` and `message=` shapes on these three at the same stroke.

Declined anyway, for one reason: three constructors **in this same module** already resolve this exact collision by overwriting, one of them landed in [#888](https://github.com/basecamp/basecamp-sdk/pull/888) days ago with the rationale written at the site. Closing the signature on three HTTP error classes while five composites keep forwarding `**kwargs` would put *two shapes for the same question* back into one module — which is, precisely, the complaint that filed this card: *"The module therefore contains both shapes."* Matching the resolution that is already there keeps one shape.

It is a real alternative and a defensible one, so it is recorded here rather than argued away. If we ever want the closed-signature shape, it should land as one change across all eight `**kwargs`-forwarding error constructors, not three of them.

**9. "The rule holds without exception" is overclaimed.** Fair. The rule governs how a constructor *treats the keyword*, not whether every class fixes the *right* retryability — and Python does not: `AuthError`, `ForbiddenError`, `NotFoundError` and `ValidationError` fix none although SPEC §6 gives their codes `retryable: false`, and `PeopleConfirmationRequiredError` is `Omit<…, "retryable">` in TypeScript and takes no such kwarg in Ruby. All of that predates this change. The claim is now qualified in the PR body rather than quietly dropped, and the divergence is worth its own look — not smuggled into this one.

---

**adversarial review: Opus** — `e1b92a01` (round 1, above) and `6376f5e5` (round 2, in flight; the head intended for merge). — @jorgemanrubia

PR #910, opened 2026-09-16
Close the malformed-trait pass in the two generators that still had it

jorgemanrubia:
[#905](https://github.com/basecamp/basecamp-sdk/pull/905) made `basecampPagination.style` decide whether a generator emits a pagination walk, and every generator refuse a value it does not implement. It said so in two shipped places — `spec/basecamp-traits.smithy` ("The six service generators reject any other value rather than reading it as 'not paginated'") and `SPEC.md` §8 ("Every service generator refuses any other value"). In two of the six that was not true.

**Kotlin gated the refusal on the object cast rather than on presence.** `operation["x-basecamp-pagination"] as? JsonObject` answers null for a trait that is present but not an object — a bare `"x-basecamp-pagination": "page"`, or a number, or an array — so `require` short-circuited true and the operation shipped as silently unpaginated.

**TypeScript gated it on truthiness.** `if (operation["x-basecamp-pagination"] && …)` exempts a present-but-falsy trait — `false`, `0`, `""` — from the refusal, same silent pass by a narrower vector.

Both are the exact failure mode #905 exists to close, and Swift already had the fix thirty lines away with a comment naming it: *"answers nil for a trait that is present but not an object, which would skip the refusal below and ship the operation as silently unpaginated."* It was written for Swift and not carried to the other two.

## The change

Presence is tested before any cast, in both, matching Swift and Rust: absent and a literal null read as unpaginated; anything else is *declared*, and a declared trait that is not an object has no style to accept, so it is refused by name. Kotlin additionally reads `style` through `as? JsonPrimitive` so a non-primitive style is refused by name rather than throwing the confusing cast error `jsonPrimitive` raises.

`parseOperation` is now exported from the TypeScript generator. #905's TypeScript tests hand-built a `ParsedOperation` and characterised `buildReturnType`, so they could not reach the parser at all — reverting the key-withholding left them green. These tests drive the real parser.

Tests land in both generators' own test modules — `kotlin/generator/src/test/…/PaginationStyleTest.kt` and `typescript/tests/generator/pagination-style.test.ts` — covering link, cursor, absent, literal null, the retired `page`, a typo, case sensitivity, a declared trait with no style, and the non-object and falsy vectors above. Kotlin's generator test module had no coverage from #905 at all.

No generated output changes: no operation declares a malformed trait, so this only alters what happens to one that would.

## Provenance

Found by an adversarial review of #905's merge commit — which arrived after #905 had already merged, because I armed auto-merge while that review was still running. That was my error and this PR is its consequence; the defects were on `main` for the interval.

The fix itself is a concurrent session's commit `a106b26c`, cherry-picked onto current `main` rather than reimplemented. It was pushed to #905's branch after that PR merged, so it was stranded where nothing would pick it up.

Raised on [Pagination trait: style "cursor" is documented but unimplemented](https://app.basecamp.com/2914079/buckets/48699913/card_tables/cards/10309068598).


jorgemanrubia:
🤖 Right, and it is this PR's own defect one level up — a test that cannot fail for the reason it is named after.

`styleFailure` already answers the exception's message, and the named-style tests assert against it. The five malformed-trait tests threw that answer away and checked only the exception type. Since `jsonPrimitive` throws `IllegalArgumentException` too, `a non-primitive style is refused by name rather than throwing a cast error` could not distinguish a refusal from the cast error in its own title.

All five now assert the message, through a helper whose comment says why the type alone is not enough.

**Proven rather than asserted.** I reverted `OperationParser.kt` to its pre-fix shape — cast-gated `require` plus `.jsonPrimitive` — and re-ran:

```
20 tests completed, 5 failed
a bare string trait is refused, not read as unpaginated()            FAILED at :42
an array trait is refused()                                          FAILED at :42
a boolean trait is refused()                                         FAILED at :42
a numeric trait is refused()                                         FAILED at :42
a non-primitive style is refused by name rather than throwing a cast error()  FAILED at :52
```

The line numbers are the point. The first four fail at `:42`, inside `assertFailsWith` — the reverted parser throws nothing at all for them, which is the silent pass this PR closes. The fifth fails at `:52`, the new message assertion — the exception *is* thrown, but it is the cast error rather than the refusal, which is precisely what the old assertion could not see. Before this commit that fifth test passed against the reverted parser.

Restored, green again: `BUILD SUCCESSFUL`, 20/20.

PR #928, opened 2026-09-18
Let the Recordings tag reach the Rust generator

jeremy:
Follow-up to #922, from a Codex review that landed on its final head a few minutes after it merged.

## The finding

#922 retagged the six recording operations `Automation` → `Recordings` and removed the now-unreachable `Automation` service splits from the five tag-keyed generators. It deliberately left `rust/generator/names.toml`'s `[operation_services]` entries alone, on the reasoning that Rust's table keys off operationId rather than the tag, so those entries were live rather than dead.

They are live. That was the problem, not the reason to keep them.

`Naming::service_for` (`rust/generator/src/naming.rs:152`) consults `operation_services` **first** and returns on a hit, reaching the tag table only when there is none. So for these six operations Rust alone kept deciding the service from a hand-written table while the other five generators followed the tag. The values agree today, which is why nothing was visibly wrong — but the next correction or retag of these operations would move five SDKs and silently leave Rust behind, which is the drift #922 exists to close.

## The change

Remove the six overrides. Rust then resolves them through the tag like everyone else.

Verified rather than assumed: `scripts/check-rust-service-drift.sh` regenerates byte-identical, and `check-operation-assignment-parity` and `check-service-inventory-parity` stay clean.

## Note on scope

This could not go into #922 — that PR had already merged when the review arrived. Filed here instead of being left as an unresolved thread on a merged PR, where nobody would have picked it up.

---
## Summary by cubic
Removes the six `Recordings` operation service overrides in the Rust generator so Rust resolves those operations through their `Recordings` tag like the other generators.

Rust consults `[operation_services]` in `rust/generator/names.toml` before falling back to the tag, so these overrides let Rust drift from the tag silently on a future retag. The values match today, so generated output stays byte-identical and no behavior changes.

<sup>Written for commit 776ec8d09aa7ec99ab60d111c71d378531e2120f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/basecamp/basecamp-sdk/pull/928?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jeremy:
**The one red check here is inherited from `main`, not caused by this change. Merge #927 first.**

`Known-defect tracking issues are open` fails on this branch with:

```
ERROR: canonical tracking claims exist outside the registry:
  - scripts/check-required-tags.rb:42 — a canonical "tracked in #925" claim, and
    [scripts/check-required-tags.rb, #925] is not registered
```

That is byte-for-byte the failure `main` itself has carried since #922 merged, and this branch was cut from `main`. #927 adds the single missing `spec/tracking-issues.yml` row; once it lands, merging `main` into this branch clears the check with no change here.

Deliberately not fixing it in this PR: both would add the same registry row and the second to merge would conflict, and a hotfix for a red `main` should not be gated behind a generator change.

The Rust change's own gates are green — `scripts/check-rust-service-drift.sh` regenerates byte-identical, and `check-operation-assignment-parity` and `check-service-inventory-parity` both pass.

PR #931, opened 2026-09-18
Find operations by what they are not, so no HTTP verb can drop one in silence

### LINKED ISSUE #925: Generators iterate a five-verb list, so an operation on any other HTTP method is silently dropped from every SDK
## The hole

Every per-language generator finds operations by iterating a hard-coded list of five HTTP verbs — `get`, `post`, `put`, `patch`, `delete`:

- `ruby/scripts/generate-services.rb:20`, `ruby/scripts/generate-metadata.rb:16`
- `python/scripts/generate_services.py:26`, `python/scripts/generate_metadata.py:33`
- `typescript/scripts/generate-services.ts:897`, `typescript/scripts/extract-metadata.ts:55`, `typescript/scripts/extract-person-id-sites.ts:67`
- `kotlin/generator/.../OperationParser.kt:304`
- `swift/Sources/BasecampGenerator/OpenAPIParser.swift:74`
- `rust/generator/src/model.rs:476`
- `scripts/generate-url-routes:66` (the jq filter), and `scripts/check-operation-assignment-parity:344`

An OpenAPI Path Item Object may carry eight operation fields in 3.1 — `get`, `put`, `post`, `delete`, `options`, `head`, `patch`, `trace` — and 3.2 adds `query` and `additionalOperations`. An operation on any verb outside the five would therefore be **dropped from the generated services entirely**: no method on any client, in six languages, silently.

## Why it is reachable rather than theoretical

Smithy's `@http` trait takes `method` as a free-form string. Per the Smithy spec, "Smithy will use this value literally and will perform no validation on the method." So a model author writing `method: "HEAD"` in `spec/basecamp.smithy` gets a valid model, a valid `openapi.json`, and no operation in any SDK. Nothing in the pipeline objects.

The model happens to use only GET/POST/PUT/DELETE today (PATCH is not used either), which is why this has never bitten.

## Related, already fixed

#922 found and fixed the same shape in `scripts/check-required-tags.rb`, which enumerated the same five verbs and so never tag-checked an `options`/`head`/`trace` operation. That gate now identifies operations by what they are *not* — the Path Item Object's non-operation fields are a closed, spec-defined set (`$ref`, `summary`, `description`, `servers`, `parameters`) plus `x-` extensions, and everything else is an operation, with an unrecognised field reported by name rather than skipped.

That guard would now at least *name* such an operation. It cannot stop the generators from dropping it.

## Why this is worth doing separately

It was deliberately scoped out of #922: fixing it means editing six generators and re-running every drift check, which does not belong in a retagging PR. But it is directly upstream of anything that builds a catalog from SDK operations — a consumer that trusts "the SDK exposes every operation" inherits this hole.

## Suggested shape

Apply the same inversion the required-tags gate now uses, rather than lengthening six lists to eight and leaving the structure that produced the bug. `scripts/check-projected-examples.rb:84` and `scripts/check-kotlin-optional-arrays-and-scalars.rb:171` already spell the eight-verb list if an enumeration is preferred in a given language, but the inversion is what stops the next verb from repeating this.

A cheaper interim: a check that fails if `openapi.json` contains a path-item operation field outside the set the generators actually handle. That converts a silent drop into a red build without touching six generators.

jeremy:
## The hole

Every per-language generator found operations by iterating its own hard-coded list of five HTTP verbs — `get`, `post`, `put`, `patch`, `delete`. Smithy's `@http` trait takes `method` as a free-form string that Smithy "will use literally and will perform no validation on", so a model author writing `method: "HEAD"` in `spec/basecamp.smithy` got a valid model, a valid `openapi.json`, and **no method on any client, in six languages, with no error**.

Measured rather than asserted. Taking the committed `openapi.json`, renaming `GetProject`'s verb from `get` to `head`, and running the **pre-change** generators:

| | operations generated | `GetProject` present | exit |
|---|---|---|---|
| Ruby, unmodified spec | 265 | yes | 0 |
| Ruby, HEAD spec | **264** | **no** | **0** |
| Python, unmodified spec | 265 | yes | 0 |
| Python, HEAD spec | **264** | **no** | **0** |

One operation gone from the SDK, and a clean exit.

## The instrument, per #922

Not a longer list. #922 fixed the same shape in `scripts/check-required-tags.rb` by identifying an operation by what it is **not**, and the same inversion lands here: a Path Item Object's non-operation fields are a closed, spec-defined set (`$ref`, `summary`, `description`, `servers`, `parameters`) plus `x-` extensions, and **every other field is an operation**. A field that is neither a known non-operation field nor a readable operation object is named, not stepped over. A `$ref` path item is refused rather than read as empty, since its operations cannot be counted without resolving it.

The verb list has already fallen behind twice — OpenAPI 3.0 added `trace`, 3.2 adds `query` and `additionalOperations` — and `@http.method` means this repo's own generator can put a key in a path item that no list anticipates.

## Loud refusal, not new emission — and the reason this matters beyond this PR

**Chosen: the generators refuse an operation on a verb they cannot emit. They were deliberately not taught to emit new verbs.**

Emitting HEAD/OPTIONS/TRACE would mean inventing transport helpers in six runtimes for a verb the Basecamp API does not serve — API surface nobody asked for, six times. Refusal costs nothing and is fail-closed: the operation cannot ship missing, because nothing ships until someone decides.

The consequence, stated because a follow-up depends on it: `scripts/gen-catalog/main.go` bounds the catalog to the generated verb set and **fails the build** on anything else. That gate stays correct and stays needed after this PR. It would have inverted into a bug only if the generators had started emitting those verbs, which they do not. Its logic is untouched here.

## One declaration, not six literals

The five verbs were restated independently in each generator, in `scripts/generate-url-routes`, and in `gen-catalog`. That copying is the mechanism, not just the symptom, so it is gone:

**`spec/generated-verbs.json`** is the single declaration, read by all six service generators. It is a **policy**, not six per-language capabilities: `check-operation-assignment-parity` already requires all six SDKs to expose the same operation set, so six independently-maintained bounds could never legitimately differ — it was one fact written six times. The array is **ordered**, and that order is emission order, so there is no second per-generator list for ordering either.

Resolution mirrors what each generator already does for `openapi.json`: a path relative to the script (Ruby, Python, TypeScript) or a `--verbs` flag defaulting beside `--openapi` (Kotlin, Swift, Rust). `gen-catalog` is peer-owned; its comment now points at the declaration, and deriving from it is theirs to take.

The file's own header carries the reasoning, including the part a declaration cannot enforce: adding an entry does not implement it. Six runtimes need the transport helper first.

## What each generator does now

| Generator | Discovery | Emission bound |
|---|---|---|
| `ruby/scripts/generate-services.rb` | inverted | declaration; refuses by name |
| `python/scripts/generate_services.py` | inverted (`gen_common.iter_operations`) | declaration; refuses by name |
| `typescript/scripts/generate-services.ts` | inverted (`scripts/path-items.ts`) | declaration; refuses by name |
| `kotlin/…/OperationParser.kt` | inverted (`PathItems.kt`) | declaration; refuses by name |
| `swift/…/OpenAPIParser.swift` | inverted (`PathItems.swift`) | declaration; refuses by name |
| `rust/generator/src/model.rs` | inverted | declaration; refuses by name |
| `ruby/scripts/generate-metadata.rb` | inverted | **none** — verb-agnostic |
| `python/scripts/generate_metadata.py` | inverted | **none** — verb-agnostic |
| `typescript/scripts/extract-metadata.ts` | inverted | **none** — verb-agnostic |
| `typescript/scripts/extract-person-id-sites.ts` | inverted | **none** — verb-agnostic |
| `scripts/generate-url-routes` (jq) | inverted | **none** — verb-agnostic |
| `scripts/check-operation-assignment-parity` | inverted | n/a — the anchor must stay total |

The four metadata extractors and the route generator key everything on `operationId` and represent a verb as data, so they take no bound and now cover an operation on **any** verb. Bounding them would have been fail-closed theatre: they can genuinely handle it.

**Go is the one generator with nothing to invert.** Its client comes from `oapi-codegen`; there is no repo-owned path-item walk. Its route table is `scripts/generate-url-routes`, which is in the table above.

## Mutation check

`scripts/test-generator-verb-inversion.rb` (new, wired into `make check` and the Spec Gates job) drives Ruby, Python and the jq route generator with crafted specs — HEAD, 3.2's `query`, a non-operation field, an unreadable field, a `$ref` path item, a declaration missing `delete`. 21 cases, all passing.

The numbers are the argument:

| walk restored to | cases failing (of 21) |
|---|---|
| the original **five**-verb list | **10** |
| the **eight** verbs OpenAPI 3.1 names | **3** |

The eight-verb list repairs HEAD and still fails on 3.2's `query` and on the two cases where a field that is neither known nor readable is stepped over in silence. That gap is why this is an inversion rather than a longer list, and why it is a test rather than a paragraph.

Case 9 and 17 are worth naming separately: they point a generator at a crafted declaration lacking `delete` and assert `DELETE` is then refused. That is what proves the bound is read from `spec/generated-verbs.json` rather than from a private literal that merely happens to match.

The compiled generators carry the identical walk; the test does not drive them (they need a JVM, a Swift toolchain, cargo or an npm install, and their drift checks run in their own CI jobs). The refusal was reproduced by hand in all four against the same HEAD spec:

```
TypeScript  exit 1  Error: openapi.json declares HEAD /{accountId}/projects/{projectId} (GetProject), and this generator emits only GET/POST/PUT/PATCH/DELETE. …
Kotlin      exit 1  (IllegalStateException, same text)
Swift       exit 1  (same text)
Rust        exit 1  (same text)
```

No partial output was written in any of them.

## Regeneration

**Byte-identical in all seven SDKs.** `openapi.json` declares four verbs and no non-operation path-item fields, so the inversion sees exactly what the lists saw. The ordering property is what makes that true and is pinned by case 8: the declaration's order is emission order, so a path item spelled `{delete, get}` still emits `get` first.

Verified green locally: the six per-language drift checks, `go-check-generated-drift`, `go-check-wrapper-drift`, `check-operation-assignment-parity`, `check-service-inventory-parity`, `check-grouped-client-coverage`, `url-routes-check`, `catalog-check`, `smithy-check`, `bc3-route-parity`, and every self-test alongside them.

Two harness fixes were needed, both because the harness stands in for the repository layout the generators resolve against:

- `scripts/check-typescript-service-drift.sh` assembles a throwaway project outside the repo; it now copies `spec/generated-verbs.json` to where `scripts/path-items.ts` resolves it, exactly as it already did for `openapi.json`.
- `rust/generator/tests/fixtures.rs` passes `--verbs` pointing at the repository's own declaration rather than a per-fixture copy — copying it into each fixture root would reintroduce, in the tests, the duplicated literal this removes.

## The tracking claims, removed in this PR because it closes #925

Two canonical `tracked in #925` claims and both `spec/tracking-issues.yml` rows are gone. Leaving either would have pointed the `Known-defect tracking issues are open` gate at a closed issue the moment this merged — red on `main` and, because it evaluates on the merge commit, on every open PR at once.

- `scripts/check-required-tags.rb` — the claim described the hole this PR closes, so the paragraph is rewritten as an as-of fact: the generators now walk by the same exclusion this gate uses. The explanation survives; only the promise is gone.
- `scripts/gen-catalog/main.go` — **comment text only, no logic change.** The consolidation half of that claim is obsolete (there is a shared source now, and taking it belongs to that generator's owner). The bound half is not obsolete and is not deleted: this PR made the generators refuse rather than emit, so a non-generated verb still reaches no client, and `gen-catalog` reads `openapi.json` directly rather than the SDK trees, so it holds that line on its own input even when nothing has been regenerated.

`make check-known-defect-issues-open` exits 0 locally, and `rg 'tracked in #925'` returns nothing.

## What this does not do

- It does not teach any SDK to serve HEAD/OPTIONS/TRACE/`query`. Those refuse, loudly.
- It does not resolve `$ref` path items anywhere. Every walk refuses one by name; resolving local references is the better end state and no walk under-counts while waiting for it.
- OpenAPI 3.2's `additionalOperations` is a map of method to Operation. Every walk here reads it as a single operation and refuses it by name. The artifact declares `3.1.0`, so it is unreachable; the failure is loud and the remedy is local to each walker.
- `spec/generated-verbs.json` declares what the generators emit; nothing mechanically verifies each runtime actually has the helper. Adding a verb there without implementing it is a compile error in TypeScript, Kotlin, Swift and Rust, and a missing-method error at call time in Ruby and Python. The file's header says to add the implementations first; that instruction is the whole of the guard, and it is stated rather than assumed.
- `scripts/check-projected-examples.rb`, `scripts/check-kotlin-optional-arrays-and-scalars.rb` and `scripts/sync-doc-constants.rb` still spell the eight-verb list. They are checks, not generators — none of them can drop an operation from an SDK — and they were out of #925's scope. Converting them is a small, separate change.

Closes #925.

---
## Summary by cubic
Fixes operations being silently dropped from generated SDKs when the model uses an HTTP verb outside the previous five-verb list. Generators now find every operation in a path item and refuse, by name, any operation they cannot emit.

**What changed**

- Path items are read by exclusion: non-operation fields and `x-` extensions are skipped; every other field is treated as an operation.
- `spec/generated-verbs.json` is now the single ordered declaration of which verbs the SDKs emit.
- Metadata extractors and the URL route generator are verb-agnostic and now cover operations on any verb.
- `$ref` path items stop generation rather than being silently ignored.

**Side effects**

- Generated SDK output is byte-identical.
- No SDK gains new verb support: HEAD, OPTIONS, TRACE, and `query` fail loudly.
- Adding a verb to `spec/generated-verbs.json` requires the corresponding transport helper in all six runtimes first.

Closes #925.

<sup>Written for commit a7f1229db3ae8186ce8b63ecb4b9aebb4813f72e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/basecamp/basecamp-sdk/pull/931?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jeremy:
@codex review

jeremy:
All eight findings from the review were real, and all eight are fixed in #932. This PR merged about nine minutes after opening — roughly forty seconds before the review landed — so none of them could be addressed here.

Per finding:

1. **Kotlin cleared its output before parsing** — confirmed and fixed. `Main.kt` deleted `models/*.kt`, `services/*.kt`, `Metadata.kt` and `ServiceAccessors.kt` before `groupOperations()`, so a refusal erased the committed tree. It parses first now, and the new compiled-generator test asserts an untouched output tree (seeded first — an empty directory cannot show data loss, which is why the first version of that test passed against the unfixed generator).
2. **`patch` is not a shared capability** — confirmed on both counts: Ruby's `BaseService` delegates only `get post put delete post_raw put_raw`, and Kotlin's `ServiceEmitter` renders PATCH through `httpPut`. Removed from `spec/generated-verbs.json`, with the two required fixes recorded there. No PATCH operation exists, so nothing regenerates differently; one now refuses by name instead of shipping broken.
3. **The compiled generators had no test that could see a regression** — correct, and the sharpest of the eight. `scripts/test-compiled-generator-refusal` runs each real binary against a HEAD spec and asserts a named refusal that writes nothing and destroys nothing; wired into `ts-check`, `kt-check`, `swift-check`, `rs-check` and each language's CI job.
4. **Kotlin's loader coerced non-strings** — fixed; every entry must now be a non-blank JSON string.
5. **Rust's `filter_map` dropped non-strings** — fixed the same way. Together these were a shared declaration that six generators read differently, which is worse than six literals.
6. **The jq filter accepted unreadable fields** — fixed; a path-item field must be an operation object with a non-empty `operationId`, so a scalar no longer errors obliquely and an unrecognised object no longer becomes a route naming `null`.
7. **Swift still `continue`d past a malformed path item** — fixed; the guard is gone and `operationsOf` validates, so the walk is total.
8. **`additionalOperations` slipped through the verb-agnostic walkers** — the most substantive. The bounded walks refused it as an unemittable verb, but the four metadata extractors and the route generator read it as one operation with no `operationId` and skipped it, dropping every operation inside. Every walk names it now.

The self-test is up from 21 cases to 28 and `make check` is green; all seven SDKs still regenerate byte-identical.

